PR #58, opened 2019-07-12
Clean up CMS navigation link previews

contolini:
Added a bare bones preview of the nav hierarchy by cheaply adding non-breaking spaces to indent the items. I also created a string widget without a preview for general use.

Screen Shot 2019-07-12 at 9 37 14 AM

### DIFF
--- a/docs/admin/src/widgets/navigation/tertiaryPreview.js
+++ b/docs/admin/src/widgets/navigation/tertiaryPreview.js
@@ -1,0 +1,5 @@
+import React from 'react';
+
+export const Preview = props => <div>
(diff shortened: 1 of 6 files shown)


docs/admin/src/widgets/navigation/tertiaryPreview.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+export const Preview = props => <div>
+    &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;└ {props.value}
```
cfarm: └33t

PR #166, opened 2019-07-26
Add Jekyll post render hook to inline icon SVGs

contolini:
Jekyll doesn't render our `{% include %}` statements because they're stored as strings in our pages' front matter. Using a Jekyll hook we can swap out our include statements with inline SVGs.

See hubcap issues #240 and #221.

Before: https://cfpb.github.io/design-system/foundation/iconography#variations (scroll down to icon table)
After: https://deploy-preview-166--cfpb-design-system.netlify.com/design-system/foundation/iconography#variations

### DIFF
--- a/docs/components/text-inputs.md
+++ b/docs/components/text-inputs.md
@@ -140,7 +140,7 @@ variations:
               title="Test input"
               class="a-text-input">
           <button class="a-btn a-btn__link">
-              <img src="/design-system/assets/icons/error.svg" class="cf-icon-svg" />
+              {% include icons/error.svg %}
               <span class="u-visually-hidden">Clear</span>
           </button>
       </div>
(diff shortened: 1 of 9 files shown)


docs/components/text-inputs.md
```diff
@@ -140,7 +140,7 @@ variations:
               title="Test input"
               class="a-text-input">
           <button class="a-btn a-btn__link">
-              <img src="/design-system/assets/icons/error.svg" class="cf-icon-svg" />
+              {% include icons/error.svg %}
```
cfarm: since this will now be the code in the HTML code snippet:
```
<a class="a-link a-link__icon" href="#">
    <span class="a-link_text">Example with icon</span>
    {% include icons/download.svg %}
</a>
```

is that ok? is this copy and pasteable like this to cfgov-refresh or some other site using capital-framework?

contolini: Correct, it's the same format used in the current CF docs:

Screen Shot 2019-07-30 at 2 54 51 PM

PR #298, opened 2019-08-28
Re-style tabs, make them responsive

contolini:
Applies CFPB styles to the GovUK tabs we're using. We may want to
formally fork their tabs and redesign them for our design system. For
now we're just tweaking their CSS to make it more CFPBy.

See hubcap/issues/291 for context.

Before:

Screen Shot 2019-08-28 at 4 42 27 AM

After:

Screen Shot 2019-08-28 at 4 42 17 AM


contolini:
@marteki @designlanguage @cfarm I updated the tabs per hubcap/issues/287.

Screen Shot 2019-08-29 at 1 33 53 PM

PR #320, opened 2019-09-06
Update specs spreadsheet and add Inches column to tables

contolini:
Adds the new "Inches" column from the spreadsheet. I also re-exported the spreadsheet.


contolini:
Closing in favor of #323.

PR #360, opened 2019-10-08
Remove cf- prefix from components and add @cfpb npm scope

contolini:
We're migrating away from the Capital Framework brand and dropping the `cf-`
prefix in favor a `@cfpb` npm scope. E.g.:

- `cf-buttons` becomes `@cfpb/buttons`
- `capital-framework` becomes `@cfpb/design-system`

As part of the migration I searched-and-replaced instances of `cf-` with `@cfpb/`.
I also updated relevant documentation but didn't look too closely at files like the
components' `usage.md` because I assume we'll be deleting them in favor of the
[design system docs](https://cfpb.github.io/design-system/components/).

I reset each component's version to 1.0.0. After we do some test publishings
and ensure everything loads and tests correctly, we can either leave the components
at 1.x.x or set them to 12.0.0, the next major bump after Capital Framework's current 11.x.x.

## Testing

- `yarn && yarn build && yarn test` should pass
- `cd docs/ && yarn && yarn start` should serve the design system website at http://localhost:4000/design-system/

## Notes

- Snyk is failing because the package manifests have changed locations. Everything else should pass.

@anselmbradford I'll hit you up to give this thing a whirl.


packages/buttons/README.md
```diff
@@ -0,0 +1,32 @@
...
+This component can be used by itself, but it was made for Design System,
+a front end framework developed at the
+[Consumer Financial Protection Bureau](https://consumerfinance.gov).
+
+## How to use this component
+
+Detailed instructions can be found at the Design System
+[documentation site](https://cfpb.github.io/design-system/).
+
+> NOTE: Be sure to run the Less files through
```
anselmbradford: I feel like this note in all the readmes is superfluous. I presume this is for when the components are imported and their source less files are compiled in a project that may have a build system that doesn't incorporate autoprefixer. Let's either move this to one note in the detailed instructions area or remove this entirely (since people could do all kinds of config things that cause issues if they're not careful, which really isn't our bailiwick).

packages/buttons/usage.md
```diff
@@ -37,7 +37,7 @@ where this component's less file is imported.
 
 ### Color variables
 
-Color variables referenced in comments are from [cf-core cf-brand-colors.less](https://github.com/cfpb/capital-framework/blob/master/packages/cf-core/src/cf-brand-colors.less).
+Color variables referenced in comments are from [@cfpb/core @cfpb/brand-colors.less](https://github.com/cfpb/design-system/blob/master/packages/core/src/brand-colors.less).
```
anselmbradford: `@cfpb/brand-colors.less` is correct here and it wouldn't be just brand-colors.less?

packages/expandables/src/expandables.less
```diff
@@ -34,9 +34,9 @@
 
 // Import external dependencies
 
-@import (less) '../../cf-core/src/cf-core.less';
-@import (less) '../../cf-buttons/src/cf-buttons.less';
-@import (less) '../../cf-icons/src/cf-icons.less';
+@import (less) '../../core/src/core.less';
+@import (less) '../../buttons/src/buttons.less';
```
anselmbradford: Can you think of any scenario where it becomes a pain in a code search to look for `buttons.less` since it's more generic?

packages/icons/usage.md
```diff
@@ -34,7 +34,7 @@ The standard icon height in ems matches the 19px rendered canvas of text
 set in Avenir Next sized at 16px (19/16 = 1.1875).
 
 ```
-@cf-icon-height: 1.1875em;
+@@cfpb/icon-height: 1.1875em;
```
anselmbradford: Oops, we don't want to be remaining variables with the scope, and the double at symbol here is definitely wrong.

test/unit-test-coverage/lcov-report/index.html
```diff
@@ -20,31 +20,31 @@ <h1>
     </h1>
     <div class='clearfix'>
       <div class='fl pad1y space-right2'>
-        <span class="strong">86.82% </span>
+        <span class="strong">Unknown% </span>
```
anselmbradford: What's going on with this coverage report. Also, the coverage report should be ignored and not committed, shouldn't it?

packages/expandables/src/expandables.less
contolini: I was thinking the same thing but it felt weird leaving `cf-` in filenames. Ideally we'd just import the node module name a la `@import '@cfpb/buttons';` and gulp would get the module's entry point based on its package.json `less` or [`style` attribute](https://jaketrent.com/post/package-json-style-attribute/). In fact, I'm surprised we never configured CF to operate like that.

packages/icons/usage.md
contolini: Bloops.

packages/expandables/src/expandables.less
anselmbradford: Sure I'm fine with `cf-` or `cfpb-`… or `bcfp-` KIDDING

package.json
```diff
@@ -29,7 +29,7 @@
     "changelog": "./scripts/generate-changelog.sh",
     "release": "yarn run build && lerna publish",
     "version": "yarn run changelog && git add CHANGELOG.md",
-    "process-icon-svgs": "svgo -o packages/icons/src/icons --enable=addClassesToSVGElement --enable=removeStyleElement --enable=removeAttrs --config='{ \"plugins\": [ { \"addClassesToSVGElement\": { \"className\": \"cf-icon-svg\" } }, { \"removeAttrs\": { \"attrs\": [ \"fill\", \"path:class\", \"circle:class\" ] } }, { \"cleanupIDs\": { \"force\": \"true\" } } ] }'"
+    "process-icon-svgs": "svgo -o packages/cfpb-icons/src/cfpb-icons --enable=addClassesToSVGElement --enable=removeStyleElement --enable=removeAttrs --config='{ \"plugins\": [ { \"addClassesToSVGElement\": { \"className\": \"cf-icon-svg\" } }, { \"removeAttrs\": { \"attrs\": [ \"fill\", \"path:class\", \"circle:class\" ] } }, { \"cleanupIDs\": { \"force\": \"true\" } } ] }'"
```
anselmbradford: In the way the scoping works shouldn't it be `packages/icons/src/cfpb-icons`?

contolini:
@anselmbradford after banging my head against the wall trying to figure out why Travis isn't passing (the tests run fine locally) I'm thinking it has something to do with this PR being opened from a fork. The [branch](https://travis-ci.com/contolini/design-system/builds/132932768) builds fine but this PR [doesn't](https://travis-ci.com/cfpb/design-system/builds/132932770).

anselmbradford:
@contolini Are you saying we should just merge it?

contolini:
I gUuEeSsSs. I'm limited in my debugging abilities. Let's do a quick pairing later today and then merge it.

PR #371, opened 2019-10-25
Remove grid-generated CSS

anselmbradford:
grid-generated CSS was a CSS-only version of the grid that could be imported into a project to get around using less. However, we don't use this ourselves and no other components support this, however, it does have build steps we need to maintain. Probably best to remove it and force a less pre-processor requirement.

## Removals

- Remove grid-generated CSS and associated build steps and documentation.


Scotchester:
I'm not strongly opposed to this, but I will point out that it will make it significantly more difficult to just grab the compiled Design System CSS and add it to a simple site with no build process to compile Less. The only multi-column layout options available would be the ones in cf-layout.

Maybe as a mitigation, we could include some very simple Less compilation guidance in the docs?

anselmbradford:
We're merging this for now as we work on publishing. I will add documentation later.

PR #370, opened 2019-10-25
Update build dependencies

anselmbradford:
## Changes

- Update `gulp-stylelint` to `10.0.0`
- Update `jest` to `24.9.0`
- Update `jest-cli` to `24.9.0`
- Update `lerna` to `3.18.3`
- Update `stylelint` to `11.1.1`
- Update `stylelint-config-standard` to `19.0.0`
- Update `terser-webpack-plugin` to `2.2.1`

## Testing

1. `gulp && gulp test` should pass.

PR #372, opened 2019-10-30
Set key for tmp path

anselmbradford:
Fixes https://github.com/cfpb/design-system/pull/370#issuecomment-548021818

## Changes

- Hardcodes a key for the temporary path variable that's used when renaming components.

## Testing

1. `gulp` should pass.


contolini:
👍 Nice fix!

PR #378, opened 2019-11-13
Simplify build

anselmbradford:
Instead of spitting out the compiled JavaScript for each component, we aggregate all component's JavaScript into the design-system JavaScript file, which simplifies the build.

Also, design-system.js.LICENSE was being built. This PR turns off extracting comments in terser to remove this (addresses https://[ghe]/CFGOV/platform/issues/3602)

## Removals

- Remove passing of component name to build and just build available components.

## Changes

- Turn off extracting comments in terser
- Add `minimize` flag, which can be flipped instead of adding/removing terser.
- Include handle-errors utility in build steps for scripts.

## Testing

1. Pull branch, and `gulp build` should spit out a minimized design-system.js file in `packages/cfpb-design-system/`


anselmbradford:
@contolini

contolini:
Looks good! I can't verify the admin section but the output and compiled docs look good. 👍

PR #405, opened 2020-01-17
cfpb-core: Remove ems from breakpoint utils

anselmbradford:
The breakpoint utility methods convert the passed in value to ems. The implementing code should decide whether to pass in pixels or ems. As is this breaks the mega menu breakpoints since the points change with custom changes to the font size.

## Changes

- Remove ems from breakpoint utility methods.

## Testing

Easiest way to see what this'll do is probably to copy these changes into cfgov-refresh.
1. Copy the contents of the media-queries.less file in this PR in the same file in node_modules/cf-core/src, or node_modules/@cfpb/cfpb-core/src if on https://github.com/cfpb/cfgov-refresh/pull/5463
2. Clean and build the source in cfgov-refresh
3. If using Chrome, go to Chrome > Preferences… and search for "Appearance" and change the font size to "Very large"
4. Visit http://localhost:8000/ and resize the page. Compare to the same behavior on https://www.consumerfinance.gov/

## Screenshots

Production:
Screen Shot 2020-01-17 at 5 51 47 PM

Localhost:
Screen Shot 2020-01-17 at 5 51 56 PM

### DIFF
--- a/packages/cfpb-core/src/media-queries.less
+++ b/packages/cfpb-core/src/media-queries.less
@@ -8,26 +8,19 @@
 //
 
 .respond-to-min( @bp, @rules ) {
-    @ems: unit( ( @bp / @base-font-size-px ), em );
-
-    @media only all and ( min-width: @ems ) {
+    @media only all and ( min-width: @bp) {
         @rules();
     }
 }
 
 .respond-to-max( @bp, @rules ) {
-    @ems: unit( ( @bp / @base-font-size-px ), em);
-
-    @media only all and ( max-width: @ems ) {
+    @media only all and ( max-width: @bp ) {
         @rules();
     }
 }
 
 .respond-to-range( @bp1, @bp2, @rules ) {
-    @ems1: unit( ( @bp1 / @base-font-size-px ), em );
-    @ems2: unit( ( @bp2 / @base-font-size-px ), em );
-
-    @media only all and ( min-width: @ems1 ) and ( max-width: @ems2 ) {
+    @media only all and ( min-width: @bp1 ) and ( max-width: @bp2 ) {
         @rules();
     }
 }


Scotchester:
Hmm… Any idea why this failed to deploy on Netlify? It's complaining about not having Yarn?

anselmbradford:
@Scotchester Yes all recent PRs in memory have been doing that. Probably something I broke along the way but with Chris out I don't know where to look

Scotchester:
> @Scotchester Yes all recent PRs in memory have been doing that. Probably something I broke along the way but with Chris out I don't know where to look

Well, not _all_ of them. I was able to review #408 that way.

anselmbradford:
> Well, not all of them. I was able to review #408 that way.

Hmm weird. There was a window where the repo was missing a `yarn.lock` file, which it looks like [netlify needs to kick off yarn](https://www.netlify.com/blog/2016/11/01/yarn-support-on-netlify/), so maybe that sent things out of whack. Any hunches you have?

anselmbradford:
@Scotchester haha ok that was the issue. I just rebased it with master and now it's good again.

Scotchester:
Nice.

And now I'm realizing that I can't test the changes in cfgov-refresh that way, of course 🤦‍♂

Scotchester:
I think I'm generally in favor of making it possible to choose other units in these mixins, but I have two thoughts:

1. Forcing them to be passed in via the same `@bp` param will necessitate significant extra overhead for doing the em calculation prior to passing the param (in the vast majority of cases where we still want to use ems). Perhaps we should add additional breakpoint vars with different units?
1. We could also let people continue to think in pixels and pass them to `@bp` by adding a new `@units` param that selects what units to use.

anselmbradford:
@Scotchester do you happen to know of an example of a breakpoint that should be ems instead of pixels?

contolini:
Yeah @Scotchester I'm curious why we decided to use ems here in the first place?

Scotchester:
@anselmbradford @contolini Sorry I keep forgetting to come back to this.

The decision in the first place was likely heavily influenced by things like the seminal 2012 article https://cloudfour.com/thinks/the-ems-have-it-proportional-media-queries-ftw/ (which, it should be noted, was updated in 2015 to note that the landscape has changed somewhat).

The broad idea is that [breakpoints should be based around when the design breaks, not on particular devices](https://www.smashingmagazine.com/2018/02/media-queries-responsive-design-2018/#don-t-target-devices-add-breakpoints-when-the-design-breaks)<sup>1</sup>, and that value is different in px if a user has zoomed or increased text size. Ironically, now that I look more closely at it, I think the example in the OP of this very PR is what we want to avoid. Is the mega menu going to look/work as expected if two of the items are breaking onto a second line?

Here's the mega menu with the changes in this PR and the browser font size set to 32px:

screenshot of overlapping text in open mega menu

I suppose we could probably alter the CSS of the mega menu to ensure that it supports the items being on two lines, but using em-based queries already solves the problem.

---

1. This is one reason why our standard breakpoints are not things like 480px, 768px, 1024px, etc., but rather semi-arbitrarily chosen numbers. We wanted the convenience of having a few common ones so you didn't have to think about a custom one ALL the time, but in specific instances we still encourage writing specific breakpoints for when the content breaks.

anselmbradford:
This was largely in service of the mega menu using pixels in its JS. https://github.com/cfpb/cfgov-refresh/pull/5613 does the same conversation in the JS that happens in the CSS to solve the issue shown in this original post's screenshots. I like the idea of being able to pass in the units to the media query, but I think for now there isn't a need this would be addressing, so I'm going to close this.

PR #406, opened 2020-01-21
Hide content when the expandable is collapsed

anselmbradford:
Adds `display: none` to the expandable's content when it is collapsed.

## Changes

- Adds placeholder links to the expandable content so they contain focusable content.
- Adds event handlers for `expandBegin` and `collapseEnd` events, which add and remove the `u-hidden` class to the expandable content DOM node.

## Testing

1. Pull branch, run `gulp build`
2. `cd docs && yarn && yarn start`
3. Visit http://localhost:4000/design-system/components/expandables
4. Using the tab key, pay attention to what has focus. When collapsed, focus should jump from the expandable header to the next focusable element on the page. When expanded, focus should jump from the expandable header, to the link in the expandable content, to the next focusable element on the page.

## Screenshots

Screen Shot 2020-01-21 at 5 12 33 PM


packages/cfpb-expandables/src/Expandable.js
```diff
@@ -47,6 +47,8 @@ function initialize() {
     this.ui.content
   );
   this.transition = transition.init();
+  this.transition.addEventListener( 'expandBegin', expandBeginHandler.bind( this ) );
```
anselmbradford: I don't believe this can go in the `events` further up in this file because it is on the transition, but if someone knows better let me know.

contolini:
@anselmbradford I think this PR may have broken the side nav on larger screens

Screen Shot 2020-02-05 at 11 59 06 AM

https://deploy-preview-406--cfpb-design-system.netlify.com/design-system/foundation/
https://cfpb.github.io/design-system/foundation/

PR #409, opened 2020-01-23
Fix package.json issues | Remove unused copy-webpack-plugin

anselmbradford:
## Changes

- Change `private` attribute in `package.json` from string to valid boolean.
- Add required `version` attribute to `package.json`.
- Remove `main` attributein `package.json` that pointed to JS file that doesn't exist.
- Move `devDependencies` next to `dependencies` in `package.json`.
- Change `homepage` attribute in `package.json` to the github pages URL, not the repo.
- Remove unused `copy-webpack-plugin`.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -2,9 +2,19 @@
   "name": "@cfpb/cfpb-design-system",
   "version": "1.0.0",
   "description": "CFPB Design System documentation website",
-  "homepage": "https://github.com/cfpb/design-system",
+  "homepage": "https://cfpb.github.io/design-system/",
   "license": "MIT",
-  "main": "dist/main.js",
+  "scripts": {
+    "start": "parallelshell \"bundle exec jekyll serve --watch --host=localhost\" \"yarn run watch\" \"yarn run watch-css\"",
+    "compile-js": "NODE_ENV=production webpack",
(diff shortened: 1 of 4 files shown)


docs/package.json
```diff
@@ -2,9 +2,19 @@
   "name": "@cfpb/cfpb-design-system",
   "version": "1.0.0",
   "description": "CFPB Design System documentation website",
-  "homepage": "https://github.com/cfpb/design-system",
+  "homepage": "https://cfpb.github.io/design-system/",
   "license": "MIT",
-  "main": "dist/main.js",
```
anselmbradford: I removed this as there isn't a `dist/main.js` file. There is a `dist/js/design-system.js` file, should it be pointing to that?

package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "cfpb-design-system",
-  "private": "true",
+  "version": "1.0.0",
```
anselmbradford: I set this to `1.0.0` as that is the same version set in `docs/package.json`.

Scotchester: Hmm... shouldn't it be `0.0.8` to match the versions of the individual component packages?

contolini: @Scotchester that was true when the project root was also the master `capital-framework` component but we now have the master component in a [sub dir](https://github.com/cfpb/design-system/tree/master/packages/cfpb-design-system) so the version number here doesn't matter.

contolini:
:shipit:

PR #425, opened 2020-02-05
Ensure collapsed expandable is visible when overflow is overridden

anselmbradford:
Fixes sidebar that was hidden from the change in https://github.com/cfpb/design-system/pull/406

## Changes

- Ensure collapsed expandable is visible when overflow is overridden

## Testing

1. Sidebar in e.g. https://cfpb.github.io/design-system/components/expandables should be visible.


contolini:
🙇

PR #438, opened 2020-04-06
Initial Design System clean up

contolini:
This is a series of small changes to let stakeholders see what the DS looks like with a more "vanilla" version of CF and Netlify CMS.

## Additions

- Global eyebrow with our standard "this is a gov website" tagline.
- Skip to main content link
- [AnchorJS](https://www.bryanbraun.com/anchorjs/) library that automatically deep links every heading.

## Removals

- Page status pills.
- Design token tables.
- "Why are we sharing this" page.

## Changes

- Pointed netlify-cms to upstream to re-add the kanban board functionality so folks can try it.
- Fixed tabs by updating the GovUK tabs library.

## Testing

1. Click around the netlify preview link below 

or

1. Pull down branch and
1. cd `docs/`
1. `yarn && yarn start`
1. Visit http://localhost:4000/design-system/

## Screenshots

Screen Shot 2020-04-06 at 12 30 29 PM


docs/assets/css/layout.less
```diff
@@ -0,0 +1,7 @@
+// taken from enhancements/layout.less
```
anselmbradford, line 1: Could add " in cfgov-refresh" as was the comment in the global header less file.

contolini: Good call.

stephanieosan:
One thing I found with this is when a live code example includes a header, that example's header becomes anchor-link-able in a kind of strange way, where it's a little confusing whether the anchor link is part of the pattern or not. Check out this screenshot, and see it in action on the [cards page](https://cfpb.github.io/design-system/components/cards#credit-cards). 
Screen Shot 2020-04-10 at 3 43 46 PM

Not exactly sure what behavior we'd prefer. On one hand, it's probably valuable to be able to link to a live example (?), but on the other, it's confusing to see a live example have an element in it that isn't actually part of its interaction pattern (the anchor link icon). 

@sonnakim @contolini @anselmbradford

anselmbradford:
Yeah the shown patterns shouldn't change, that's just a recipe for confusion over how the example should actually function. Link should only be on the heading for the example:

Screen Shot 2020-04-10 at 4 03 33 PM

PR #446, opened 2020-04-15
Scaffold out some browser tests

contolini:
Sets up webdriver.io with a couple simple spec files. Consider this an experiment -- if we think this is a good idea and we write more tests we should integrate it into our `yarn test` workflow.

## Additions

- Two commands: `yarn webdriver` and `yarn webdriver-sauce`
- Two webdriver config files, one for local use, one for sauce.
- Two spec files, one that opens/closes an expandable and another that verifies the hide/show code toggle button works correctly on the docs website (because I keep breaking this feature).

## Testing

1. Pull down branch.
1. `yarn`
1. Ensure the docs website is running locally if it isn't already (`cd docs/ && yarn start`)
1. `yarn webdriver` to run the tests using chrome locally.
1. `yarn webdriver-sauce` to run the tests against four different browsers in sauce labs. Note: you'll need to define `SAUCE_USERNAME` and `SAUCE_ACCESS_KEY` env variables.

## Notes

- Running the specs through sauce shows that the hide/show code button doesn't work in IE11, I believe because babel isn't properly transpiling the `docs/assets/js/main.js` file.


test/browser/packages/cfpb-expandables.js
```diff
@@ -0,0 +1,34 @@
...
+    isExpandableContentVisible = () => expandableContent.isDisplayed();
+  } );
+
+  it( 'should hide expandable content on page load', () => {
+    expect( isExpandableContentVisible() ).toBeFalsy();
+  } );
+
+  it( 'should open expandable when header is clicked', () => {
+    expandableHeader.click();
+    browser.pause( 300 );
```
anselmbradford, line 24: Instead of a pause, would a wait work, like https://webdriver.io/docs/api/element/waitForDisplayed.html

test/browser/docs/code-snippets.js
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable no-undef */
+
+describe( 'the code snippet toggling feature', () => {
```
anselmbradford, line 3: Trivial: top-level `describe` should capitalize the first word `The`

test/browser/packages/cfpb-expandables.js
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable no-undef */
+
+describe( 'basic cfpb expandable', () => {
```
anselmbradford, line 3: Trivial: top-level `describe` should capitalize the first word `Basic`

package.json
```diff
@@ -30,13 +30,22 @@
     "changelog": "./scripts/generate-changelog.sh",
     "release": "yarn run build && lerna publish",
     "version": "yarn run changelog && git add CHANGELOG.md",
-    "process-icon-svgs": "svgo -o packages/cfpb-icons/src/cfpb-icons --enable=addClassesToSVGElement --enable=removeStyleElement --enable=removeAttrs --config='{ \"plugins\": [ { \"addClassesToSVGElement\": { \"className\": \"cf-icon-svg\" } }, { \"removeAttrs\": { \"attrs\": [ \"fill\", \"path:class\", \"circle:class\" ] } }, { \"cleanupIDs\": { \"force\": \"true\" } } ] }'"
+    "process-icon-svgs": "svgo -o packages/cfpb-icons/src/cfpb-icons --enable=addClassesToSVGElement --enable=removeStyleElement --enable=removeAttrs --config='{ \"plugins\": [ { \"addClassesToSVGElement\": { \"className\": \"cf-icon-svg\" } }, { \"removeAttrs\": { \"attrs\": [ \"fill\", \"path:class\", \"circle:class\" ] } }, { \"cleanupIDs\": { \"force\": \"true\" } } ] }'",
+    "webdriver": "wdio run test/browser/webdriver.conf.js",
```
anselmbradford, line 34: Ideally we should separate the command from the implementation, so we might think about giving this a more general name like in cfgov-refresh, like `test-acceptance`

package.json
```diff
@@ -30,13 +30,22 @@
...
+    "webdriver-sauce": ": ${SAUCE_USERNAME?\"You need to define SAUCE_USERNAME and SAUCE_ACCESS_KEY env variables \"} && wdio run test/browser/webdriver-sauce.conf.js"
```
anselmbradford, line 35: …and maybe `test-acceptance-cloud`

test/browser/packages/cfpb-expandables.js
Scotchester: And `CFPB` 😝

test/browser/webdriver.conf.js
```diff
@@ -0,0 +1,265 @@
...
+  //
+  capabilities: [ {
+
+    // maxInstances can get overwritten per capability. So if you have an in-house Selenium
+    // grid with only 5 firefox instances available you can make sure that not more than
+    // 5 instances get started at a time.
+    maxInstances: 5,
+    //
+    browserName: 'chrome'
+    // If outputDir is provided WebdriverIO can capture driver session logs
```
anselmbradford, line 56: Let's make the local test headless! You can add this here:
```
'goog:chromeOptions': {
     // to run chrome headless the following flags are required
     // (see https://developers.google.com/web/updates/2017/04/headless-chrome)
     args: ['--headless', '--disable-gpu'],
 }
```

test/browser/webdriver.conf.js
```diff
@@ -0,0 +1,265 @@
...
+  // https://docs.saucelabs.com/reference/platforms-configurator
+  //
+  capabilities: [ {
+
+    // maxInstances can get overwritten per capability. So if you have an in-house Selenium
+    // grid with only 5 firefox instances available you can make sure that not more than
+    // 5 instances get started at a time.
+    maxInstances: 5,
+    //
+    browserName: 'chrome'
```
anselmbradford, line 55: Let's make the local test headless! You can add this here:
```
browserName: 'chrome',
'goog:chromeOptions': {
     // to run chrome headless the following flags are required
     // (see https://developers.google.com/web/updates/2017/04/headless-chrome)
     args: ['--headless', '--disable-gpu'],
 }
```

contolini: Nice!

package.json
contolini: How about `test:browser` for now? Until we figure out exactly what we're going to do with webdriver (unit, integration, acceptance)

anselmbradford: That's better than webdriver I think and seems to be what they get called in the team. It's still a bit implementation-centric though. As an analogous example, we wouldn't call the unit tests "test:cli" or something because they run on the command-line. That's my only point, if there's something that describes the type of test, not where it runs, that would be better.

PR #456, opened 2020-04-23
Make homepage editable via the CMS

contolini:
Adds netlify [file collection](https://www.netlifycms.org/docs/collection-types/#file-collections)  called 'Generic pages' that is used for one-off pages with unique layouts. Currently, the homepage is the only item in this collection. A jinja template and netlify preview template was created to unintelligently render whatever HTML and markdown is added to a generic page's "body" field.

## Additions

- Netlify CMS file collection w/ homepage file.
- Template for rendering and previewing "generic content", i.e. a freeform text field containing HTML and/or markdown.
- Edit button on homepage.

## Testing

1. Compare the [production homepage](https://cfpb.github.io/design-system/) to this [PR preview homepage](https://deploy-preview-456--cfpb-design-system.netlify.app/design-system). They should be the same.
1. Click the edit button in the bottom right and type random HTML and markdown into the text field. It should render in the preview pain. (Don't click save or it'll open a PR). Note: the text field will be blank when the page loads because Netlify CMS pulls its content from the master branch of this repo and the [home.md](https://github.com/cfpb/design-system/blob/master/docs/home.md) file doesn't have a `body` field until this PR is merged.

## Screenshots

Screen Shot 2020-04-23 at 5 56 39 PM

### DIFF
--- a/docs/_includes/generic-content.html
+++ b/docs/_includes/generic-content.html
@@ -1,0 +1,9 @@
+<section class="content_main">
+
+    {% if page.body %}
+    <section>
+        {{ page.body | markdownify }}
+    </section>
+    {% endif %}
+
+</section><!-- .content_main -->
--- a/docs/_includes/generic-content.html
+++ b/docs/_includes/generic-content.html
@@ -1,9 +1,9 @@
 <section class="content_main">
 
     {% if page.body %}
-    <section>
+    <div>
         {{ page.body | markdownify }}
-    </section>
+    </div>
     {% endif %}
 
 </section><!-- .content_main -->
--- a/docs/admin/src/netlify-cms.js
+++ b/docs/admin/src/netlify-cms.js
@@ -1,5 +1,6 @@
 import CMS from 'netlify-cms';
 import { default as componentsPreviewTemplate } from './widgets/componentsPreviewTemplate';
+import { default as genericPreviewTemplate } from './widgets/genericPreviewTemplate';
 import { Preview as navPreview } from './widgets/navigationPreviewTemplate';
 
 CMS.registerPreviewTemplate( 'getting-started', componentsPreviewTemplate );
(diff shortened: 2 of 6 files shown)


docs/_includes/generic-content.html
```diff
@@ -0,0 +1,9 @@
+<section class="content_main">
```
anselmbradford, line 1: Does this have an element with a `content` class as a parent container to this? It should.

docs/_includes/generic-content.html
```diff
@@ -0,0 +1,9 @@
...
+
+    {% if page.body %}
+    <section>
```
anselmbradford, line 4: Maaaybe this is a `div` since its direct parent is a section with nothing else in it but this element?

docs/admin/src/netlify-cms.js
```diff
@@ -1,12 +1,13 @@
...
 
 CMS.registerPreviewTemplate( 'getting-started', componentsPreviewTemplate );
 CMS.registerPreviewTemplate( 'foundation', componentsPreviewTemplate );
 CMS.registerPreviewTemplate( 'components', componentsPreviewTemplate );
 CMS.registerPreviewTemplate( 'templates', componentsPreviewTemplate );
 CMS.registerPreviewTemplate( 'navigation', navPreview );
-
+CMS.registerPreviewTemplate( 'home', genericPreviewTemplate );
 
 CMS.registerPreviewStyle( '/design-system/dist/css/main.css' );
```
anselmbradford, line 13: Not a change in this PR, but seems like we should set up a variable for the JavaScript paths equivalent to the `{{ site.baseurl }}` variable for asset paths used in the templates.

docs/admin/src/widgets/genericPreviewTemplate.js
```diff
@@ -0,0 +1,23 @@
+import React, { Component } from 'react';
+import { ReactLiquid, liquidEngine } from 'react-liquid';
+import marked from 'marked';
+import template from '../../../_includes/generic-content.html';
```
anselmbradford, line 4: 🤔 this path

contolini: Hmmm yeah maybe I'll move that filepath prefix into a const called something like `JEKYLL_TEMPLATE_LOCATION` to clarify why it needs to reach into such a distant directory.

docs/_includes/generic-content.html
contolini: Eww yeah works for me.

PR #467, opened 2020-04-29
Remove unused redirects plugin

chosak:
We're not using this plugin right now, so removing it until such time as we want or need to support redirects this way.


anselmbradford:
This LGTM, but I leave it to Chris' for approval if this is somehow used somewhere. Like this doesn't tie into the [permalinks](https://github.com/cfpb/design-system/blob/master/docs/_plugins/permalinks.rb) plugin somehow, does it?

PR #470, opened 2020-04-30
Add link to pages' GitHub history in footer

contolini:
## Removes

- All the CSS rules in `footer.less` that aren't used anywhere. 

## Changes

- Links pages' "Last edited" timestamp to their GitHub history pages.

## Testing

1. Check out the PR preview link below.

## Screenshots

Screen Shot 2020-04-30 at 6 20 54 PM

### DIFF
--- a/docs/_includes/footer.html
+++ b/docs/_includes/footer.html
@@ -7,9 +7,14 @@
         <div class="block block__sub">
             <small>
                 <strong>Page last edited:</strong>
-                <time datetime="{{ page.last_modified_at }}">
-                    {{ page.last_modified_at | date: "%l:%M:%S %p %b %d, %Y" }}
-                </time>
+                {% if page.section %}
+                    {% assign file_url_section = page.section | append: '/' %}
+                {% endif %}
+                <a href="https://github.com/{{ site.repository }}/commits/master/docs/{{ file_url_section }}{{ page.path | split: '/' | last }}">
+                    <time datetime="{{ page.last_modified_at }}">
+                        {{ page.last_modified_at | date: "%l:%M:%S %p %b %d, %Y" }}
+                    </time>
+                </a>
             </small>
         </div>
         {% endif %}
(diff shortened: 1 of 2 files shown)


docs/_includes/footer.html
```diff
@@ -7,9 +7,14 @@
         <div class="block block__sub">
             <small>
                 <strong>Page last edited:</strong>
-                <time datetime="{{ page.last_modified_at }}">
-                    {{ page.last_modified_at | date: "%l:%M:%S %p %b %d, %Y" }}
-                </time>
+                {% if page.section %}
```
contolini, line 10: This conditional is here because not all pages (like the homepage) have a page section.

anselmbradford:
This is cool. One thing I'm wondering about. Say I visit the blocks page, https://deploy-preview-470--cfpb-design-system.netlify.app/design-system/components/blocks, go to the footer and click "5:34:12 PM Apr 10, 2020", the top commit is dated today (May 1), not Apr 10. Is that expected? https://github.com/cfpb/design-system/commits/master/docs/components/blocks.md

contolini:
@anselmbradford I think that's because this PR preview was built yesterday and that latest commit entered master a couple hours ago. Let me rebuild the preview and let's see if it updates.

docs/_includes/footer.html
anselmbradford: So within this `if` block you assign `file_url_section`, but then you use it outside the `if` block on L:13. The `if` should enclose the link too, no? Or provide a non-link version?

contolini: Yeah it only gets defined if needed. Otherwise, it's undefined and prints nothing on L13. I can change the logic if it's confusing.

anselmbradford: Maybe this?

```
{% assign file_url_path = page.path | split: '/' | last %}
{% if page.section %}
    {% assign file_url_path = page.section | append: '/' | concat: file_url_path %}
{% endif %}
<a href="https://github.com/{{ site.repository }}/commits/master/docs/{{ file_url_path }}">
```

contolini: Works for me 👍

PR #474, opened 2020-05-01
Add support for optional local backend

chosak:
Add support for and document running against a local GitHub proxy server instead of against the live branch on GitHub.com.

This is a beta feature of Netlify CMS:

https://www.netlifycms.org/docs/beta-features/#working-with-a-local-git-repository

I've pushed this commit to master on my fork, and you can see that [my fork's GH pages](https://chosak.github.io/design-system/) edit buttons still try to authenticate to GitHub.com; but if you [run this branch locally](http://localhost:4000/design-system/), it will use the local proxy server, but only if is running. If you're not running a local proxy server, local development still falls back to the GitHub backend.

Per the netlify-cms code, the CMS only looks for the local proxy [if running on localhost](https://github.com/netlify/netlify-cms/issues/2335#issuecomment-615168281).

I don't _love_ the recommendation of `npx` to run random code from npm. Is there a better way to do this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,14 +75,8 @@ yarn start
 ```
 
 Visit http://localhost:4000/design-system/ to view the site.
-Editing a page locally will open a real PR against this GitHub repository.
-
-Alternatively, you can run a local proxy server that mimics GitHub, using
+Editing a page locally will change your local files, using
 [an experimental Netlify CMS feature](https://www.netlifycms.org/docs/beta-features/#working-with-a-local-git-repository).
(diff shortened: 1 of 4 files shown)


chosak:
@contolini thanks, that's a great improvement! It runs locally for me using just `yarn start`, very nicely (although required me upgrading my local Node to the latest 10.x, 10.10.1). Would you mind giving another quick look at how I've merged your changes?

contolini:
Yeeeee :shipit:

PR #475, opened 2020-05-01
Fix variations search | add search fields | highlight search term

anselmbradford:
## Changes

- Fix missing variations field search by including new variations_group parent.
- Add `research` and `related_items` fields to search.
- Highlight search term in results with `<mark>` element.

## Testing

1. Perform a search on PR preview.

## Screenshots

Screen Shot 2020-05-01 at 6 12 35 PM

### DIFF
--- a/docs/_includes/search-scripts.html
+++ b/docs/_includes/search-scripts.html
@@ -16,7 +16,7 @@
                 {% for variation in variation_group.variations %}
                     'variation_name{{ forloop.index }}': '{{ variation.variation_name | xml_escape }}',
                     'variation_description{{ forloop.index }}': {{ variation.variation_description | markdownify | strip_html | strip_newlines | jsonify }},
-                    'variation_code_snippet{{ forloop.index }}': {{ variation.variation_code_snippet | markdownify | strip_html | strip_newlines | jsonify }},
+                    'variation_code_snippet{{ forloop.index }}': `{{ variation.variation_code_snippet | xml_escape }}`,
                 {% endfor %}
             {% endfor %}
 
--- a/docs/_includes/search-scripts.html
+++ b/docs/_includes/search-scripts.html
@@ -5,25 +5,29 @@
     window.searchStore = {
         {% for page in site.pages %}
         '{{ page.url | slugify }}': {
+
+            // Cycle through the variation fields of this page.
+            {% for variation_group in page.variation_groups %}
+                // Determine the maximum number of variations present overall.
+                {% if variation_group.variations.size > variations_count %}
+                    {% assign variations_count = variation_group.variations | size  %}
+                {% endif %}
+
+                {% for variation in variation_group.variations %}
+                    'variation_name{{ forloop.index }}': '{{ variation.variation_name | xml_escape }}',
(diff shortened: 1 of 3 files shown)


docs/_includes/search-scripts.html
```diff
@@ -5,25 +5,29 @@
...
+            {% for variation_group in page.variation_groups %}
+                // Determine the maximum number of variations present overall.
+                {% if variation_group.variations.size > variations_count %}
+                    {% assign variations_count = variation_group.variations | size  %}
+                {% endif %}
+
+                {% for variation in variation_group.variations %}
+                    'variation_name{{ forloop.index }}': '{{ variation.variation_name | xml_escape }}',
+                    'variation_description{{ forloop.index }}': {{ variation.variation_description | markdownify | strip_html | strip_newlines | jsonify }},
+                    'variation_code_snippet{{ forloop.index }}': {{ variation.variation_code_snippet | markdownify | strip_html | strip_newlines | jsonify }},
```
contolini, line 19: The `strip_html` filter results in many of the code snippet fields being empty on the search results page. For example, searching for `block__flush` doesn't return any results because the CSS classes get removed as part of `strip_html`.

Is there a safe way to include the HTML in the code snippets? Could we wrap `variation.variation_code_snippet` in backticks (<code>```</code>) and then pass it to `markdownify`, resulting in all the HTML/CSS surviving but being safely escaped?

docs/assets/js/search.js
```diff
@@ -23,24 +22,19 @@ function displaySearchResults( results, store ) {
...
-        if ( typeof itemFieldValue !== 'undefined' && itemFieldValue !== '' ) {
-          previewText = itemFieldValue;
-          break;
-        }
-      }
+      // Remove fields that should never appear as the preview.
+      delete searchMatchFields.id;
+      delete searchMatchFields.title;
+
+      previewText = item[ Object.keys( searchMatchFields )[0] ];
```
contolini, line 32: My linter informs me there shouldn't be a space after this bracket. Now that GitHub Actions is configured for this repo we should add a PR linting job.

docs/assets/js/search.js
```diff
@@ -23,24 +22,19 @@ function displaySearchResults( results, store ) {
...
+
+      const regex = new RegExp( results.searchTerm, 'gi');
+      previewText = previewText.replace( regex, function replace( match ) {
+        return '<mark>' + match + '</mark>';
```
contolini, line 36: When you demo this at stand-up get the team's opinion on the default bright yellow color for `mark`. And 💯 for the semantic HTML.

docs/assets/js/search.js
anselmbradford: Fixed! 🦊

docs/assets/js/search.js
anselmbradford: They saw. They said nothing.

docs/_includes/search-scripts.html
anselmbradford: I `xml_escaped` it and put it in a template string.

PR #479, opened 2020-05-04
Search refinements

anselmbradford:
## Changes

- Exclude search page itself from search results.
- Omit preview text if there aren't any fields to pull from (e.g. search for "homepage").
- Provide a result count and term at the top of search results.
- Provide search term in no results display.
- Don't float the search input box on mobile.

## Testing

1. Check preview and search for "homepage", "search", "block__flush", etc. and resize to mobile.

## Screenshots
Screen Shot 2020-05-04 at 5 49 20 PM

### DIFF
--- a/docs/assets/js/search.js
+++ b/docs/assets/js/search.js
@@ -63,7 +63,7 @@ function displaySearchResults( elm, results, store ) {
  * @param {string} term - the search term
  */
 function displayNoSearchResults( elm, term ) {
-  elm.innerHTML = `<li>No results found for '${ term }'.</li>`;
+  elm.innerHTML = `<li>No search results found for '${ term }'.</li>`;
 }
 
 /**
--- a/docs/assets/js/search.js
+++ b/docs/assets/js/search.js
@@ -9,43 +9,56 @@ const searchResultsElm = document.getElementById( 'search-results' );
  * Update page markup with search results.
  * @param {Array} results - A list of search result hits as objects.
  * @param {Object} store - search index/meta data store in the window object.
+ * @returns {boolean} True if there are search results, false otherwise.
  */
 function displaySearchResults( results, store ) {
   // Are there any results?
-  if ( results.length ) {
-    let appendString = '';
+  if ( results.length === 0 ) {
+    displayNoSearchResults( searchResultsElm, results.searchTerm );
+    return false;
+  }
 
-    // Iterate over the results.
-    for ( let i = 0; i < results.length; i++ ) {
-      const item = store[results[i].ref];
-      appendString += '<li><a href="../' + item.url + '"><h3>' + item.title + '</h3></a>';
+  // Search results content will be placed in here.
+  let resultsString = `
+    <p>
+      ${ results.length } result${ results.length > 1 ? 's' : '' }
+      for search of '${ results.searchTerm }'.
+    </p>
+  `;
 
-      // Show some preview text under each search results item.
-      let previewText = '';
-      const searchMatchWordFragment = Object.keys( results[i].matchData.metadata )[0];
-      const searchMatchFields = results[i].matchData.metadata[searchMatchWordFragment];
+  // Iterate over the results.
+  for ( let i = 0; i < results.length; i++ ) {
+    const item = store[results[i].ref];
+    resultsString += '<li><a href="../' + item.url + '"><h3>' + item.title + '</h3></a>';
 
-      // Remove fields that should never appear as the preview.
-      delete searchMatchFields.id;
-      delete searchMatchFields.title;
+    // Show some preview text under each search results item.
+    let previewText = '';
+    const searchMatchWordFragment = Object.keys( results[i].matchData.metadata )[0];
+    const searchMatchFields = results[i].matchData.metadata[searchMatchWordFragment];
 
-      previewText = item[Object.keys( searchMatchFields )[0]];
+    // Remove fields that should never appear as the preview.
+    delete searchMatchFields.id;
+    delete searchMatchFields.title;
 
-      const regex = new RegExp( results.searchTerm, 'gi' );
-      previewText = previewText.replace( regex, function replace( match ) {
-        return '<mark>' + match + '</mark>';
-      } );
-
-      // Add the preview text.
-      if ( previewText !== '' ) {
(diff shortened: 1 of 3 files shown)


docs/assets/js/search.js
```diff
@@ -112,4 +134,6 @@ if ( searchTerm ) {
 
   // Display the results of the search.
   displaySearchResults( results, searchStore );
+} else {
```
contolini, line 137: Kill this `else` condition and in `search-results.html` add something like `<li>Enter search terms above.</li>` in the unordered list. It'll be removed if search results are found.

docs/assets/js/search.js
```diff
@@ -9,43 +9,56 @@ const searchResultsElm = document.getElementById( 'search-results' );
  * Update page markup with search results.
  * @param {Array} results - A list of search result hits as objects.
  * @param {Object} store - search index/meta data store in the window object.
+ * @returns {boolean} True if there are search results, false otherwise.
```
contolini, line 12: Curious: what's the value in returning a boolean? Is there a future use case for it?

docs/assets/js/search.js
```diff
@@ -9,43 +9,56 @@ const searchResultsElm = document.getElementById( 'search-results' );
...
-      const searchMatchWordFragment = Object.keys( results[i].matchData.metadata )[0];
-      const searchMatchFields = results[i].matchData.metadata[searchMatchWordFragment];
-
-      // Remove fields that should never appear as the preview.
-      delete searchMatchFields.id;
-      delete searchMatchFields.title;
+  if ( results.length === 0 ) {
+    displayNoSearchResults( searchResultsElm, results.searchTerm );
+    return false;
+  }
```
contolini, line 19:
```suggestion
```

Because `results` is known before we invoke `displaySearchResults`, I'd prefer to avoid calling the function altogether instead of returning early. See below.

docs/assets/js/search.js
```diff
@@ -9,43 +9,56 @@ const searchResultsElm = document.getElementById( 'search-results' );
...
+      return '<mark>' + match + '</mark>';
+    } );
 
-      // Add the preview text.
-      if ( previewText !== '' ) {
-        appendString += '<p>' + previewText.substring( 0, 150 ) + '…</p></li>';
-      }
+    // Add the preview text.
+    if ( previewText !== '' ) {
+      resultsString += '<p>' + previewText.substring( 0, 150 ) + '…</p></li>';
```
contolini, line 55: Is this a use case for `text-overflow: ellipsis`?

docs/assets/js/search.js
```diff
@@ -112,4 +134,6 @@ if ( searchTerm ) {
 
   // Display the results of the search.
   displaySearchResults( results, searchStore );
```
contolini, line 136: w/r/t L19

```suggestion
  if ( results.length > 0 ) {
    displaySearchResults( results, searchStore );
  }
```

docs/assets/js/search.js
```diff
@@ -9,43 +9,56 @@ const searchResultsElm = document.getElementById( 'search-results' );
...
+    <p>
+      ${ results.length } result${ results.length > 1 ? 's' : '' }
+      for search of '${ results.searchTerm }'.
+    </p>
+  `;
+
+  // Iterate over the results.
+  for ( let i = 0; i < results.length; i++ ) {
+    const item = store[results[i].ref];
+    resultsString += '<li><a href="../' + item.url + '"><h3>' + item.title + '</h3></a>';
```
contolini, line 32: If you're using a template literal above on L22 maybe use it every time you do string interpolation for consistency's sake.

docs/assets/js/search.js
anselmbradford: Won't that show `<li>Enter search terms above.</li>` to the user while the search is in progress potentially if it ends up taking seconds to execute like 18f's does? Also, if there are no search results I was putting in the search term, which is why it was written via JS. Do you think that's too much?

Screen Shot 2020-05-07 at 4 36 28 PM

docs/assets/js/search.js
anselmbradford: It was purely in service of the guard clause that returned the function early. That could also return undefined instead.

docs/assets/js/search.js
contolini: Aw yeah, you're right. Disregard my comments, it works great how it is.

docs/assets/js/search.js
anselmbradford: Maaaaybe. So here's the trade-off. With `text-overflow` all the preview text would be the same width, which might look nice. It will be only one line though, because to get `text-overflow: ellipsis` to work I have to set the text to not wrap. Also, with `text-overflow`, all the text gets written to the DOM (unless we substring it at some intermediate amount) and then it gets hidden by the CSS, so it's probably more intensive in re-draw. Which do you prefer?

(A)

Screen Shot 2020-05-08 at 10 16 07 AM

(B)

Screen Shot 2020-05-08 at 10 16 57 AM

docs/assets/js/search.js
anselmbradford: Updated in https://github.com/cfpb/design-system/pull/479/commits/2f47506c0040baa2cf8c924d7b750300f8ee560e

docs/assets/js/search.js
anselmbradford: Chat in MM is to keep is (B). An advantage with that is we need no CSS.

docs/assets/js/search.js
anselmbradford: Refactored away in https://github.com/cfpb/design-system/pull/479/commits/04489eac20472c1a1c293054c09e98dcee1c100b

docs/assets/js/search.js
anselmbradford: Refactored in https://github.com/cfpb/design-system/pull/479/commits/04489eac20472c1a1c293054c09e98dcee1c100b

docs/assets/js/search.js
anselmbradford: Refactored in https://github.com/cfpb/design-system/pull/479/commits/04489eac20472c1a1c293054c09e98dcee1c100b

PR #481, opened 2020-05-06
Replace Travis CI tests with GitHub Actions workflow

contolini:
Adds two jobs -- one for unit tests and one for browser tests piped through sauce labs. I separated them because the browser tests take a while to spin up and complete.

## Additions

- GitHub Actions workflow file for tests.

## Removals

- Travis config file.

## Changes

- Safari on Mac OS 10.15 in sauce labs is flaky (I noticed it'll sometimes hang) so I downgraded it to 10.14.

## Testing

- Check the results of the two jobs in the status check section below.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,57 @@
...
+
+    - name: Start local Design System website server
+      run: |
+        yarn --cwd docs/ start &
+        sleep 5
+
+    - name: Run browser tests
+      env:
+        SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
+        SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
```
contolini, line 56: I added these secrets to this repo's [settings page](https://github.com/cfpb/design-system/settings/secrets). The access key is tied to my CFPB sauce labs account.

anselmbradford: Can we leave a single line comment (should be supported in yml with a hash) that says this?

test/browser/webdriver-sauce.conf.js
```diff
@@ -54,7 +54,7 @@ exports.config = {
     {
       browserName: 'safari',
       browserVersion: 'latest',
-      platformName: 'macOS 10.15'
+      platformName: 'macOS 10.14'
```
anselmbradford, line 57: Why is the downgrade necessary?

PR #482, opened 2020-05-06
Update dependencies

anselmbradford:
Follow up to https://github.com/cfpb/design-system/pull/471 but updates mostly dependencies in root folder to align with /docs/ folder.

## Testing

1. Pull branch, run yarn, and yarn start. It should work.


anselmbradford:
@contolini I'm going to merge this as there're a lot of moving parts and this should be up-to-date with the earlier update deps PR. I'll fix it if it breaks something.

PR #487, opened 2020-05-11
Convert restrictions do/do not to use icon

anselmbradford:
## Changes

- Convert restrictions do/do not section to use an icon instead of a bar.

## Testing

1. Check buttons page on the PR preview, or any other page that has a restrictions section.

## Screenshots

Before:
Screen Shot 2020-05-11 at 12 32 26 PM

After:
Screen Shot 2020-05-11 at 12 32 17 PM

## Notes

- The restrictions section should just be deleted, really

### DIFF
--- a/docs/assets/css/variation.less
+++ b/docs/assets/css/variation.less
@@ -53,3 +53,12 @@
         background-color: @white;
     }
 }
+
+//
+.component-restrictions_heading-do svg {
+    color: @green;
+}
+
+.component-restrictions_heading-donot svg {
+    color: @red;
+}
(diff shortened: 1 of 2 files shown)


docs/assets/css/variation.less
```diff
@@ -53,3 +53,12 @@
         background-color: @white;
     }
 }
+
+//
```
contolini, line 57: Empty comment.

anselmbradford: lol oops

anselmbradford: Added and fixed.

PR #489, opened 2020-05-11
cfpb-layout: Support hr in content_line

anselmbradford:
The content line usually appears as `<div class="content_line"></div>`, yet semantically this sure seems like it should be a horizontal rule (`<hr>`) element? When you make it a horizontal rule it is too thick as the border needs to be removed.

## Removals

- Removes redundant `content_line` class from design system buttons docs page.

## Changes

- Converts `content_line` to `hr`.
- Adds no border and margin to support `content_line` on an `hr` element.

## Testing

1. Check buttons and layout page in this PR preview and inspect the source to see where the `content_line` is. It should be the same as the appearance on production.

### DIFF
--- a/docs/_includes/variation-content.html
+++ b/docs/_includes/variation-content.html
@@ -159,12 +159,12 @@
     {% if page.restrictions != empty and page.restrictions or page.usage and page.usage != '' %}
         <section class="sticky-header u-mb45">
             <h2 id="usage">Usage</h2>
-            <div class="content_line__wide content_line content_line u-mb20"></div>
-                {% if page.usage and page.usage != '' %}
-                    <p>
-                    {{ page.usage | markdownify }}
-                    </p>
-                {% endif %}
+            <hr class="content_line content_line__wide u-mb20">
+            {% if page.usage and page.usage != '' %}
+                <p>
+                {{ page.usage | markdownify }}
+                </p>
+            {% endif %}
             {% if page.restrictions and page.restrictions != empty %}
                 <div class="content-l component-restrictions">
                     <div class="content-l_col content-l_col-1">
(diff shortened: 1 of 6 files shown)


Scotchester:
> The content line usually appears as `<div class="content_line"></div>`, yet semantically this sure seems like it should be a horizontal rule (`<hr>`) element? When you make it a horizontal rule it is too thick as the border needs to be removed.

I think I disagree with the premise.

> The HTML `<hr>` element represents a thematic break between paragraph-level elements: for example, a change of scene in a story, or a shift of topic within a section.

– https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr

That's not how `content_line` was originally conceived (though it may be used in some places that way now). `content_line` mainly was used to divide whole sections, like between header and body areas.

I think the preferred steps to take regarding these things would be:
1. The visual styling that `content_line` provides would be applied to the element directly preceding or following it, and `content_line` gets eliminated entirely.
2. We give `<hr>` some sensible default styling, which we have needed for a long time. It shouldn't need a class.

(Some of the things using `content_line` currently would probably need to be changed to plain ol' `<hr>` with the new sensible styling.)

Scotchester:
Addendum: I think the division between the header and main content areas of the "sidebar breakout" template was the driving use case behind `content_line`.

anselmbradford:
I agree with all that Scott. I guess this PR is a question of whether in the current implementation should support being in an `hr` or not. MDN has this too:

> Historically, this has been presented as a horizontal rule or line. While it may still be displayed as a horizontal rule in visual browsers, this element is now defined in semantic terms, rather than presentational terms, so if you wish to draw a horizontal line, you should do so using appropriate CSS.

I'm fine closing it if we don't want extra CSS in there as a stop-gap. The empty divs were just weird.

Scotchester:
I've had this bee in my bonnet for a while, so if you're not in a hurry, let me do some testing in cfgov-refresh and see how much of a pain it would be to go straight to the desired destination.

anselmbradford:
Cool, fine with me.

PR #492, opened 2020-05-12
Break usage field into use_cases, content_guidelines, and behavior

anselmbradford:
## Changes

- Break `usage` field into `use_cases`, `content_guidelines`, and `behavior` fields.

## Testing

1. Check PR preview and see that pages have new fields where once they had only `usage`!


docs/_includes/variation-content.html
```diff
@@ -220,7 +238,7 @@ <h2 id="accessibility">Accessibility</h2>
     {% if page.research and page.research != '' %}
         <section class="sticky-header u-mb45">
             <h2 id="research">Research</h2>
-            <div class="content_line__wide content_line content_line u-mb20"></div>
```
contolini, line 223: 🤦

PR #493, opened 2020-05-12
Inject SVG icons before rendering preview in Netlify CMS

contolini:
A little hacky but it works and I'm unsure of a better way. I added a browser test to catch future regressions.

## Changes

- The variation template strips out any icon include statements before rendering.

## Testing

1. `yarn test:browser` should pass locally.
1. Editing the buttons pages on the [PR preview](https://deploy-preview-493--cfpb-design-system.netlify.app/design-system/admin/#/collections/components/entries/buttons) should correctly show the preview pane while on [production](https://cfpb.github.io/design-system/admin/#/collections/components/entries/buttons) it throws an error and is blank.

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/docs/admin/src/widgets/variationPreviewTemplate.js
+++ b/docs/admin/src/widgets/variationPreviewTemplate.js
@@ -5,6 +5,10 @@ import { Tabs } from 'govuk-frontend';
 import marked from 'marked';
 import template from '../../../_includes/variation-content.html';
 
+// react-liquid (https://github.com/aquibm/react-liquid/) isn't able to `include` other files so we
+// strip out any instances of {% include icons/XXXXX.svg %}
+const cleanTemplate = template.replace( /{%\s+include\s+\/?icons\/([\w-]+)\.svg\s+%}/g, '' );
+
 export default class Preview extends Component {
 
   componentDidMount() {
@@ -30,7 +34,7 @@ export default class Preview extends Component {
     };
     return (
       <div>
-        <ReactLiquid template={template} data={data} html />
+        <ReactLiquid template={cleanTemplate} data={data} html />
       </div>
     );
   }
--- a/docs/admin/src/widgets/variationPreviewTemplate.js
+++ b/docs/admin/src/widgets/variationPreviewTemplate.js
@@ -6,8 +6,11 @@ import marked from 'marked';
 import template from '../../../_includes/variation-content.html';
 
 // react-liquid (https://github.com/aquibm/react-liquid/) isn't able to `include` other files so we
-// strip out any instances of {% include icons/XXXXX.svg %}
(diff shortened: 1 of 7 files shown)


docs/admin/src/widgets/variationPreviewTemplate.js
```diff
@@ -5,6 +5,10 @@ import { Tabs } from 'govuk-frontend';
 import marked from 'marked';
 import template from '../../../_includes/variation-content.html';
 
+// react-liquid (https://github.com/aquibm/react-liquid/) isn't able to `include` other files so we
+// strip out any instances of {% include icons/XXXXX.svg %}
+const cleanTemplate = template.replace( /{%\s+include\s+\/?icons\/([\w-]+)\.svg\s+%}/g, '' );
```
anselmbradford, line 10: If you're doing a find and replace, why don't we find and replace the icon file name with the icon SVG?

Scotchester:
I'm not 100% sure I understand the problem this PR is trying to solve, but our preferred method of including SVGs is having their raw SVG code be present in the final page markup. Can we do that here instead of the `img src` method?

docs/admin/src/widgets/variationPreviewTemplate.js
contolini: Good thinking! I pointed them to the SVG files.

contolini:
> I'm not 100% sure I understand the problem this PR is trying to solve, but our preferred method of including SVGs is having their raw SVG code be present in the final page markup. Can we do that here instead of the `img src` method?

This fix is just for the Netlify CMS preview pane that has to convert our liquid templates into React components. The final page markup is compiled via Jekyll and has the SVGs inlined.

Scotchester:
> This fix is just for the Netlify CMS preview pane that has to convert our liquid templates into React components. The final page markup is compiled via Jekyll and has the SVGs inlined.

Cool, all good with me, then.

contolini:
Update: @anselmbradford had the 🔥 idea of using webpack to inject the SVGs so now we're doing that instead of `img src`.

Note: `{% include icons/XXXXX.svg %}` statements that users add to a form field as part of a page's content, like in the restrictions section, do NOT get transpiled. This is because the preview pane is rendered on every keystroke and it'd be really expensive (thus, laggy) if we iterated through every field and swapped in SVGs in real time as people typed. Instead, in this PR, we just do it once when we load the static template file.

PR #494, opened 2020-05-13
Design System: Simplify markup and make two breakpoints for header and footer

anselmbradford:
Strips out some of the nesting markup for the header and footer and directly adds margin/padding. The content area uses the `content` layout classes so still has a tablet size (larger padding than mobile).

## Changes

- Makes two breakpoints for the header and footer.

## Testing

1. See PR preview.

## Screenshots

Mobile: 
Screen Shot 2020-05-21 at 1 02 31 PM

Tablet:
Screen Shot 2020-05-21 at 1 02 41 PM

Desktop:
Screen Shot 2020-05-21 at 1 02 52 PM


anselmbradford:
Hold this as I'm going to get rid of the excess breakpoints first.

anselmbradford:
Ready for review

docs/assets/css/header.less
```diff
@@ -29,88 +23,101 @@
...
 
-.respond-to-max( @bp-xs-max, {
+    .site-header-wrapper {
+        padding-bottom: unit( 30px / @base-font-size-px, em );
+        border-bottom: 4px solid @green;
+    }
 
     .site-header {
-        background-color: @gray-5;
+        max-width: 1180px;
```
contolini, line 85: Should we be using breakpoint or grid variables for values like this? Or is this `max-width` super specific to this element?

anselmbradford: So we could tie it to [`@grid_wrapper-width`](https://github.com/cfpb/design-system/blob/f4613641062df12454f3101e09232a7b94096f0e/packages/cfpb-grid/src/cfpb-grid.less#L15) and do the math to downgrade it to the values here. I dunno, what do you think? On the one hand I'd err on the side of not coupling the website so much with the DS variables. I think the DS website should be bare minimum independent CSS as much as possible so that the website for displaying the components doesn't break when the components are broken. On the other hand, these values are here to align them with the content area, which does use the DS layout classes and variables.

contolini: > I think the DS website should be bare minimum independent CSS as much as possible so that the website for displaying the components doesn't break when the components are broken.

I def agree with this statement so I vote for whatever is most congruent with it.

PR #502, opened 2020-05-18
Fix "show details" functionality in CMS

contolini:
Fixes the "show details" link within Netlify CMS's live preview pane. The Gov UK tabs we use modifies the URL's hash which doesn't play well with Netlify CMS so a simpler version of the tab switching is provided for the CMS.

## Additions

- Simplified version of the Gov UK tab switching code for the CMS to use.
- Browser test that clicks some "show details" tabs in the CMS.
- Slugify liquid filter for the preview pane (before this PR, the `slugify` filter erroneously wasn't being recognized)

## Changes

- Moves the "show details" and tabs code into their own files so they can be imported as needed.

## Testing

1. `yarn test:browser` should pass locally
1. Editing a page with variation details (like buttons) via this PR's preview link should show functioning tabs while [production](https://cfpb.github.io/design-system/admin/#/collections/components/entries/buttons) does not.

## Screenshots

cms-tabs

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/docs/admin/src/widgets/variationPreviewTemplate.js
+++ b/docs/admin/src/widgets/variationPreviewTemplate.js
@@ -1,7 +1,7 @@
 const { AllHtmlEntities } = require( 'html-entities' );
 import React, { Component } from 'react';
 import { ReactLiquid, liquidEngine } from 'react-liquid';
-import { changeTab, initTabs } from '../../../assets/js/tabs.js';
+import { changeTab, init as initTabs } from '../../../assets/js/tabs.js';
 import ReactDOM from 'react-dom';
 import marked from 'marked';
 import slugify from 'slugify';
@@ -14,11 +14,6 @@ const templateWithIcons = template.replace(
   /{%\s+include\s+\/?icons\/([\w-]+)\.svg\s+%}/g,
   ( match, icon ) => require( `../../../_includes/icons/${ icon }.svg` )
 );
-
-/**
- * @param {MouseEvent} event - The mouse event object from the click.
- */
-
 
 export default class Preview extends Component {
 
@@ -33,31 +28,26 @@ export default class Preview extends Component {
     this.containerRef = React.createRef();
   }
 
+  /**
+   * @param {MouseEvent} event - The mouse event object from the click.
+   */
   handleClick( event ) {
     const target = event.target;
-    // Returns the preview component's native DOM element for
-    // use with our non-React vanilla JS toggling and tabs functionality.
-    // https://reactjs.org/docs/react-dom.html#finddomnode
-    const doc = ReactDOM.findDOMNode( this );
     if ( target.matches( '[data-toggle-code]' ) ) {
       event.preventDefault();
-      toggleDetails( target, doc );
+      toggleDetails( target, this.containerRef.current );
     }
     if ( target.matches( '.govuk-tabs__tab' ) ) {
       event.preventDefault();
-      changeTab( target, doc );
+      changeTab( target, this.containerRef.current );
     }
   }
 
   componentDidMount() {
-    const container = ReactDOM.findDOMNode( this.containerRef.current );
-    container.addEventListener( 'click', this.handleClick );
+    const container = this.containerRef.current;
     // The Gov UK tab styles require a parent element to have .js-enabled, see tabs.less
     container.classList.add( 'js-enabled' );
-    // Initialize the Gov UK tabs once everything has finished loading
-    window.addEventListener( 'load', () => {
-      initTabs( container );
-    } );
+    initTabs( container );
   }
 
   render() {
@@ -65,7 +55,7 @@ export default class Preview extends Component {
       page: this.props.entry.toJS().data
     };
     return (
-      <div ref={this.containerRef}>
+      <div ref={this.containerRef} onClick={event => this.handleClick( event )}>
         <ReactLiquid template={templateWithIcons} data={data} html />
       </div>
     );
--- a/docs/admin/src/widgets/variationPreviewTemplate.js
+++ b/docs/admin/src/widgets/variationPreviewTemplate.js
@@ -1,9 +1,12 @@
 const { AllHtmlEntities } = require( 'html-entities' );
 import React, { Component } from 'react';
 import { ReactLiquid, liquidEngine } from 'react-liquid';
-import { Tabs } from 'govuk-frontend';
+import { changeTab, initTabs } from '../../../assets/js/tabs.js';
+import ReactDOM from 'react-dom';
 import marked from 'marked';
+import slugify from 'slugify';
 import template from '../../../_includes/variation-content.html';
+import toggleDetails from '../../../assets/js/toggle-details.js';
 
 // react-liquid (https://github.com/aquibm/react-liquid/) isn't able to `include` other files so we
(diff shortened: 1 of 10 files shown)


docs/admin/src/widgets/variationPreviewTemplate.js
```diff
@@ -12,31 +15,57 @@ const templateWithIcons = template.replace(
   ( match, icon ) => require( `../../../_includes/icons/${ icon }.svg` )
 );
 
+/**
+ * @param {MouseEvent} event - The mouse event object from the click.
```
anselmbradford, line 19: This orphan jsdoc looks like it was associated with a function that was deleted or moved.

docs/admin/src/widgets/variationPreviewTemplate.js
```diff
@@ -12,31 +15,57 @@ const templateWithIcons = template.replace(
...
 
-    if ( tabs && tabs.length > 0 ) {
-      document.querySelector( 'iframe' ).contentWindow.document.body.classList.add( 'js-enabled' );
-      for ( let i = 0; i < tabs.length; i++ ) {
-        const tab = tabs[i];
-        new Tabs( tab ).init( );
-      }
+    this.containerRef = React.createRef();
+  }
+
```
anselmbradford, line 35: jsdoc from top of the file goes here, ye?

docs/admin/src/widgets/variationPreviewTemplate.js
```diff
@@ -12,31 +15,57 @@ const templateWithIcons = template.replace(
...
+  handleClick( event ) {
+    const target = event.target;
+    // Returns the preview component's native DOM element for
+    // use with our non-React vanilla JS toggling and tabs functionality.
+    // https://reactjs.org/docs/react-dom.html#finddomnode
+    const doc = ReactDOM.findDOMNode( this );
```
anselmbradford, line 41: Do we need to look this up on every click, or are the DOM references known and can be stored?

docs/assets/js/main.js
```diff
@@ -29,29 +30,14 @@ if ( tabs && tabs.length > 0 ) {
...
-  const codeEl = document.querySelector( target.getAttribute( 'href' ) );
-  if ( codeEl && codeEl.classList.contains( HIDDEN_CLASS ) ) {
-    codeEl.classList.remove( HIDDEN_CLASS );
-    container.querySelector( '[data-toggle-code="hide"]' ).classList.remove( HIDDEN_CLASS );
-    container.querySelector( '[data-toggle-code="show"]' ).classList.add( HIDDEN_CLASS );
-  } else {
-    codeEl.classList.add( HIDDEN_CLASS );
-    container.querySelector( '[data-toggle-code="hide"]' ).classList.add( HIDDEN_CLASS );
-    container.querySelector( '[data-toggle-code="show"]' ).classList.remove( HIDDEN_CLASS );
+  if ( target.matches( '[data-toggle-code]' ) ) {
```
anselmbradford, line 38: Would it make sense to confine all references to `[data-toggle-code]` to the same module? Right now there's a reference to it in `main.js`, `toggle-details.js`, and `variationPreviewTemplate.js`. Maybe it's looked up in toggle-details, which stores DOM references that the other modules then reference.

docs/assets/js/tabs.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @param {DOMNode} container - Parent element containing Gov UK tab HTML
+ */
+export function init( container ) {
+  const tabs = container.querySelectorAll( '[data-module="tabs"]' );
+
+  if ( tabs && tabs.length > 0 ) {
+    for ( let i = 0; i < tabs.length; i++ ) {
+      const tab = tabs[i];
+      new Tabs( tab ).init( );
```
anselmbradford, line 12: Run `gulp lint --fix` if you haven't. I think it may strip out the empty space in the method call.

docs/assets/js/tabs.js
```diff
@@ -0,0 +1,51 @@
...
+ * The original Gov UK tabs code modifies the URL's hash which breaks Netlify
+ * so we replace their functionality with some simple tab switching below.
+ *
+ * @param {DOMNode} tab - Tab element that was clicked
+ * @param {DOMNode} document - Defaults to window.document but overridable for ReactDOM
+ */
+export function changeTab( tab, document = window.document ) {
+  const TABS_CONTAINER_CLASS = '.govuk-tabs';
+  const TAB_CLASS = '.govuk-tabs__list-item';
+  const TAB_CLASS_SELECTED = 'govuk-tabs__list-item--selected';
```
anselmbradford, line 28: Mix of including the leading `.` and not is potentially confusing. I would say leave off the leading `.` in all of them, as that matches how the `classList` API expects class names. Then where you need the period, have `.${ TAB_CLASS }` and where you don't have `TAB_CLASS_SELECTED `.

docs/assets/js/toggle-details.js
```diff
@@ -0,0 +1,19 @@
...
+  const container = button.parentNode;
+  const codeEl = document.querySelector( button.getAttribute( 'href' ) );
+  if ( codeEl && codeEl.classList.contains( HIDDEN_CLASS ) ) {
+    codeEl.classList.remove( HIDDEN_CLASS );
+    container.querySelector( '[data-toggle-code="hide"]' ).classList.remove( HIDDEN_CLASS );
+    container.querySelector( '[data-toggle-code="show"]' ).classList.add( HIDDEN_CLASS );
+  } else {
+    codeEl.classList.add( HIDDEN_CLASS );
+    container.querySelector( '[data-toggle-code="hide"]' ).classList.add( HIDDEN_CLASS );
+    container.querySelector( '[data-toggle-code="show"]' ).classList.remove( HIDDEN_CLASS );
```
anselmbradford, line 17: We should probably store these DOM references (either at page load or on first click of the toggle) instead of looking them up on each click of the toggle.

docs/admin/src/widgets/variationPreviewTemplate.js
contolini: On second thought, I think we can get rid of it and just use the react container ref.

docs/assets/js/main.js
contolini: I like this line of thought. It's tricky because we use the functionality in the CMS and in the static site which are very different environments. I think for now I'll just export the attribute selector `data-toggle-code` so that it's only defined in one place and the separate files can import it.

docs/assets/js/tabs.js
contolini: Done in https://github.com/cfpb/design-system/pull/502/commits/11e47e89371308e03d646e9c95bb4d50f2b8db6e

docs/assets/js/tabs.js
contolini: Done in https://github.com/cfpb/design-system/pull/502/commits/698ef41b0fb01e4c36ac8654d3be0c847bca7ec4

PR #543, opened 2020-05-26
Trim whitespace from Netlify CMS string widgets

contolini:
I copied the [default string widget](https://github.com/netlify/netlify-cms/blob/c28cc0c9e7c7bc4bed07c02dfb869b2dedab9aab/packages/netlify-cms-widget-string/src/StringControl.js) and tweaked it to strip whitespace. The `identifier_field` property was added to prevent the CMS from complaining about the `title` field [not being a string](https://www.netlifycms.org/docs/collection-types/#folder-collections):

> Folder collections must have at least one field with the name title for creating new entry slugs. That field should use the default string widget. If you wish to use a different field as your identifier, set identifier_field to the field name.

Also see on how to add a custom widget: https://www.netlifycms.org/docs/custom-widgets/

## Additions

- `StringTrimmed` Netlify CMS widget

## Changes

- All the fields that were `string` are now `string-trimmed`

## Testing

1. `yarn test:browser` should pass locally.
1. Edit a page locally and put a bunch of spaces at the start of the title field. Click `Publish now`. It will update the relevant markdown file on your local machine and the `title` field in that markdown file shouldn't have any spaces at the beginning of it.

## Notes

-

## Todos

-

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


docs/admin/config.yml
```diff
@@ -469,10 +472,11 @@ collections:
     create: true
     delete: true
     slug: '{{slug}}'
+    identifier_field: title
```
anselmbradford, line 475: Do we need this? Isn't `title` the default and you'd set `identifier_field` if you were using a different field?

contolini: If the title field isn't a `string`, the cms throws an error. This is because they're worried about people using something like a `markdown` field which would be difficult to convert into an identifier string. The `identifier_field` assuages the cms' anxiety and affirms that our `string-trimmed` widget is safe.

contolini:
> One question. Otherwise works. Are there any maintenance concerns of having this custom component? Will this need to be synced up with updates to the `string` component periodically?

Nah by registering a new widget we're not tied to the default `string` widget. It'd be bad if we were modifying the upstream source code but we're not.

PR #548, opened 2020-06-01
Fix component template rendering logic

contolini:
We were erroneously checking for only component restrictions and use cases when rendering component pages, resulting in editors having to put dummy content in the use cases field to get anything to appear. We should check for all four sections: use cases, content guidelines, behavior and restrictions.

Fixes: GHE/CFPB/el-camino/issues/238

## Changes

- Check for the existence of `page.content_guidelines` and `page.behavior` along with restrictions and use cases.

## Testing

1. The [PR preview](https://deploy-preview-548--cfpb-design-system.netlify.app/design-system/components/headings#use-cases) headings page should not have a Use Cases section while  [production](https://cfpb.github.io/design-system/components/headings#use-cases) has a placeholder period in that section.

### DIFF
--- a/docs/components/text-inputs.md
+++ b/docs/components/text-inputs.md
@@ -171,7 +171,7 @@ variation_groups:
               <textarea class="a-text-input a-text-input__full"
                         id="full-textarea-example">Placeholder text</textarea>
           </div>
-use_cases: .nan
+use_cases: ''
 content_guidelines: >-
   Choose the appropriate width for text input fields so they match the kind of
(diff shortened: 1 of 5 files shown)


docs/components/text-inputs.md
```diff
@@ -171,7 +171,7 @@ variation_groups:
               <textarea class="a-text-input a-text-input__full"
                         id="full-textarea-example">Placeholder text</textarea>
           </div>
-use_cases: .nan
```
anselmbradford, line 174: :wut:

PR #599, opened 2020-06-15
Search: Convert xml_escape filters to escape to handle single quotes

anselmbradford:
Search was breaking when content contained single quotes. This changes the filter used to escape single quotes.

## Changes

- Change `xml_escape` to `escape` filter.

## Testing

1. On this PR's preview branch search for "image" and see that there are search results.

### DIFF
--- a/docs/_includes/search-scripts.html
+++ b/docs/_includes/search-scripts.html
@@ -18,14 +18,14 @@
                 {% endif %}
 
                 {% for variation in variation_group.variations %}
-                    'variation_name{{ forloop.index }}': '{{ variation.variation_name | xml_escape }}',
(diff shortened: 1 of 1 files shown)


anselmbradford:
> Unrelated: Is it intentional for some values to be wrapped in single quotes and others like the variation_descriptions to not be wrapped in quotes? It works fine because Jekyll automatically wraps the naked values in double quotes but we might want to standardize all the values.

This crept in from how [this article](https://learn.cloudcannon.com/jekyll/jekyll-search-using-lunr-js/) formatted things, which I used as the base for building the search. I thought it was something about how `xml_escape` worked and it needed to be in quotes, but with this change to `escape` we should standardize it if it doesn't break anything.

PR #673, opened 2020-06-29
Don't strip newlines from search index

chosak:
Stripping newlines like this means that text in headings gets erroneously combined with subsequent lines. For example, try this search:

https://cfpb.github.io/design-system/search/?searchQuery=coherent

This returns no results, incorrectly. Now try this one:

https://cfpb.github.io/design-system/search/?searchQuery=reaching.coherent

This search shouldn't work.

This change removes the use of `strip_newlines`.

Fixes el-camino#264.

### DIFF
--- a/docs/_includes/search-scripts.html
+++ b/docs/_includes/search-scripts.html
@@ -19,20 +19,20 @@
 
                 {% for variation in variation_group.variations %}
                     'variation_name{{ forloop.index }}': '{{ variation.variation_name | escape }}',
-                    'variation_description{{ forloop.index }}': {{ variation.variation_description | markdownify | strip_html | strip_newlines | jsonify }},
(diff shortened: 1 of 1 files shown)


chosak:
The [search for "coherent"](https://deploy-preview-673--cfpb-design-system.netlify.app/design-system/search/?searchQuery=coherent) works on the preview environment. The browser test failures seem unrelated to this PR.

anselmbradford:
Another thought is whether we convert newlines to a space?

chosak:
Thanks for the review @anselmbradford.

> Another thought is whether we convert newlines to a space?

I couldn't find any references to this. It doesn't seem like it would make a difference either way. Let's go with this and see if we have any problems.

PR #677, opened 2020-07-01
Add spacing when long lists of jump links wrap

chosak:
Thanks to @Scotchester for the suggestion.

To test, visit a page like https://cfpb.github.io/design-system/components/buttons in a narrow browser, and observe how the buttons wrap.

|Current|With this change|
|-|-|
|image|image|

### DIFF
--- a/docs/_includes/variation-content.html
+++ b/docs/_includes/variation-content.html
@@ -40,7 +40,7 @@
     {% endfor %}
 
     {%- if jump_links %}
-    <ul class="m-list m-list__horizontal u-mb15">
+    <ul class="m-list m-list__horizontal m-list__horizontal-spaced">
         {% for jump_link in jump_links %}
         <li class="m-list_item">
             <a class="m-list_link" href="#{{ jump_link | slugify }}">
--- a/docs/_includes/variation-content.html
+++ b/docs/_includes/variation-content.html
@@ -40,7 +40,7 @@
     {% endfor %}
 
     {%- if jump_links %}
-    <ul class="m-list m-list__horizontal u-mb30">
+    <ul class="m-list m-list__horizontal u-mb15">
         {% for jump_link in jump_links %}
         <li class="m-list_item">
             <a class="m-list_link" href="#{{ jump_link | slugify }}">
--- a/docs/assets/css/variation.less
+++ b/docs/assets/css/variation.less
@@ -69,6 +69,7 @@
     color: @red;
 }
 
-.m-list__horizontal > .m-list_item {
+.m-list__horizontal-spaced,
+.m-list__horizontal-spaced > .m-list_item {
     margin-bottom: unit( 15 / @base-font-size-px, em );
 }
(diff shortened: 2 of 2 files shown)


anselmbradford:
We have a spaced list modifier [here](https://cfpb.github.io/design-system/components/lists#spaced-list), but it likely doesn't satisfy what you're after. There are two options to tighten up this change:

OPTION A: a new modifier

For this option we create a new modifier, maybe `m-list__spaced-items` or similar, and the CSS would be:

```
.m-list__spaced-items,
.m-list__spaced-items > .m-list_item {
    margin-bottom: unit( 15 / @base-font-size-px, em );
}
```

And the markup would be:

```
<ul class="m-list m-list__horizontal m-list__spaced-items">
```

OPTION B: augment existing modifier

It appears the `m-list__spaced` modifier is intended for vertical lists, not horizontal, so you could make your adjustments when both the `__horizontal` and `__spaced` modifiers are present.

```
.m-list__horizontal.m-list__spaced {
    &,
    .m-list_item + .m-list_item {
        // Negate existing spaced styles
        margin-top: 0 !important;
    }

    &,
    & > .m-list_item {
         margin-bottom: unit( 15 / @base-font-size-px, em );
    }
}
```

And the markup would be:

```
<ul class="m-list m-list__horizontal m-list__spaced">
```


__Hopefully those both work. I worked this out in my head, not code__

chosak:
@anselmbradford I've implemented option A. See live at https://deploy-preview-677--cfpb-design-system.netlify.app/design-system/components/buttons.

docs/_includes/variation-content.html
```diff
@@ -40,7 +40,7 @@ <h1 class="component-header u-mb30">
     {% endfor %}
 
     {%- if jump_links %}
-    <ul class="m-list m-list__horizontal u-mb30">
+    <ul class="m-list m-list__horizontal m-list__horizontal-spaced">
```
anselmbradford, line 43: Oops, I see one mistake in the tweaks I gave. You originally had `u-mb30` here, which gives 30px of bottom margin. `m-list__horizontal-spaced` gives 15px of bottom margin. So if 15px is fine then we're all good, but if not, you'd change the CSS to (also probably would make sense to include the `px`):

```
.m-list__horizontal-spaced {
    margin-bottom: unit( 30px / @base-font-size-px, em );
}

.m-list__horizontal-spaced > .m-list_item {
    margin-bottom: unit( 15px / @base-font-size-px, em );
}
```

chosak: I defer to you and the designers (@sonnakim) about this. It looks like the original implementation from @contolini had 30px ([here](https://github.com/cfpb/design-system/blame/99625d8c0c838b6bc71097d798756fa667152bd0/docs/_includes/variation-content.html#L30)), so maybe we should stick with that?

chosak: I've pushed a commit that changes this to 30px.

chosak:
> if this is an official promoted pattern.

Right, I wasn't sure if we wanted this to become official or not.

PR #684, opened 2020-07-02
Add banner when redirected from CF/DM

chosak:
When visitors are redirected from the soon-to-be-deprecated [Capital Framework](https://cfpb.github.io/capital-framework/) and [Design Manual](https://cfpb.github.io/design-manual/) websites, the landing URLs will have certain query string parameters:

?utm_medium=redirect&utm_source=capitalframework
?utm_medium=redirect&utm_source=designmanual

When this traffic comes in, we want to show a banner to users to let them know they've been redirected.

This change adds that banner using JavaScript.

## Screenshots

When visiting http://localhost:4000/design-system/?utm_source=capitalframework&utm_medium=redirect:

image

When visiting http://localhost:4000/design-system/?utm_source=designmanual&utm_medium=redirect:

image

## Notes

I'd very much welcome feedback on the JS implementation of this. I wasn't sure if there was some better way to do the templating. Also seeking feedback on the text content of the banners.

## Checklist

### DIFF
--- a/docs/assets/js/main.js
+++ b/docs/assets/js/main.js
@@ -42,3 +42,41 @@ function handleDocumentClick( event ) {
 }
 
 document.addEventListener( 'click', handleDocumentClick, false );
+
+// Show redirect banner if we're coming from the now-deprecated Capital
+// Framework or Design Manual websites.
+if ( window.location.search.match( /[?&]utm_medium=redirect([&#]|$)/ ) ) {
+    const match = window.location.search.match( /[?&]utm_source=([^&#]*)/ );
+
+    const redirectSources = {
+        capitalframework: {
+            name: 'Capital Framework',
+            url: 'https://cfpb.github.io/capital-framework-archive/'
+        },
+        designmanual: {
+            name: 'the CFPB Design Manual',
+            url: 'https://cfpb.github.io/design-manual-archive/'
+        }
+    };
+
+    if ( match ) {
+        const source = redirectSources[ match[ 1 ] ];
+
+        if ( source ) {
+            const banner = document.querySelector( '#redirect-banner' );
+
+            banner.querySelectorAll( 'span.redirect-source-name' ).forEach(
+                name => { name.textContent = source.name; }
+            );
+
+            banner.querySelectorAll( 'a.redirect-archive-website' ).forEach(
+                link => {
+                    link.textContent = source.url;
+                    link.href = source.url;
+                }
+            );
+
+            banner.classList.remove( 'u-hidden' );
+        }
+    }
+}
--- a/docs/assets/js/main.js
+++ b/docs/assets/js/main.js
@@ -64,17 +64,17 @@ if ( window.location.search.match( /[?&]utm_medium=redirect([&#]|$)/ ) ) {
 
         if ( source ) {
             const banner = document.querySelector( '#redirect-banner' );
+            const sourceNames = banner.querySelectorAll( 'span.redirect-source-name' );
+            const links = banner.querySelectorAll( 'a.redirect-archive-website' );
 
-            banner.querySelectorAll( 'span.redirect-source-name' ).forEach(
-                name => { name.textContent = source.name; }
-            );
+            for ( let i = 0, len = sourceNames.length; i < len; i++ ) {
+                sourceNames[i].textContent = source.name;
+            }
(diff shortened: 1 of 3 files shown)


chosak:
To test, visit:

- https://deploy-preview-684--cfpb-design-system.netlify.app/design-system/?utm_source=capitalframework&utm_medium=redirect
- https://deploy-preview-684--cfpb-design-system.netlify.app/design-system/?utm_source=designmanual&utm_medium=redirect

docs/assets/js/main.js
```diff
@@ -42,3 +42,41 @@ function handleDocumentClick( event ) {
 }
 
 document.addEventListener( 'click', handleDocumentClick, false );
+
+// Show redirect banner if we're coming from the now-deprecated Capital
+// Framework or Design Manual websites.
+if ( window.location.search.match( /[?&]utm_medium=redirect([&#]|$)/ ) ) {
+    const match = window.location.search.match( /[?&]utm_source=([^&#]*)/ );
+
+    const redirectSources = {
```
anselmbradford, line 51: This could be moved into the `if ( match ) {` block below, since it's not used outside of that.

docs/assets/js/main.js
```diff
@@ -42,3 +42,41 @@ function handleDocumentClick( event ) {
...
+        }
+    };
+
+    if ( match ) {
+        const source = redirectSources[ match[ 1 ] ];
+
+        if ( source ) {
+            const banner = document.querySelector( '#redirect-banner' );
+
+            banner.querySelectorAll( 'span.redirect-source-name' ).forEach(
```
anselmbradford, line 68: `forEach` is not supported in IE11, if that matters.

anselmbradford:
Is there a maintenance burden here? Just to play devil's advocate, what if we redirect without a banner?

docs/assets/js/main.js
chosak: It's on our supported browsers list, isn't it? I used it [over here](https://github.com/cfpb/cfgov-refresh/blame/594f148c37c691ab61450aca766af70725e32413/cfgov/unprocessed/js/modules/youtube-api.js#L34) on the video code refactor. Does that need to be changed as well? What's a more compatible alternative?

chosak:
> Is there a maintenance burden here? Just to play devil's advocate, what if we redirect without a banner?

See internal el-camino#255 for previous discussion. Personally I think the banner is nice because it'd be pretty unexpected for users coming from DM/CF links to be dropped on the DS homepage without some other warning.

docs/assets/js/main.js
anselmbradford: I just checked in Sauce Labs and yeah the banner just won't show up in IE11. You can use a regular for loop for max compatibility:

```
const sourceNames = banner.querySelectorAll( 'span.redirect-source-name' );

for ( let i = 0, len = sourceNames.length; i < len; i++ ) {
  sourceNames[i].textContent = source.name;
}
```

Also, for the video code refactor it may be okay since that code goes through Babel. But yeah that should be double-checked. Sorry I missed commenting on it there.

chosak: @anselmbradford I checked the video code in Sauce in IE11 and it seems to work:

image

So our use of Babel somehow transpiles the code so that it's compatible? And thus it doesn't need to be fixed over there? I'll make your suggested change here.

chosak: @anselmbradford I've pushed a commit that fixes the banner in IE11, using your suggestion:
image

anselmbradford: > So our use of Babel somehow transpiles the code so that it's compatible? And thus it doesn't need to be fixed over there? I'll make your suggested change here.

Yeah the idea with Babel is you write in modern JS and then give it a list of supported browsers and it transpiles the code to something that works. However, we don't polyfill things to reduce size, so it's possible some features may not work if they require a polyfill. Often it's better to re-write the code versus including a hefty polyfill. You could convert the video code to a plain for loop with little overhead if you wanted.

anselmbradford: Instead of `_callbacks.forEach( cb => cb() );`

You'd have:
```
for ( let i = 0, len = _callbacks.length; i < len; i++ ) {
  _callbacks[i]();
}
```

PR #774, opened 2020-07-08
Re-work secondary navigation

anselmbradford:
## Changes

- Exchange frankenstein secondary expandable with a details/summary element.
- Rework side nav structure to have `ds-nav-1`, `ds-nav-2`, `ds-nav-3` tree hierarchy of classes. 
- Remove `content` layout classes.
- Changes the breakpoints to be desktop/tablet and mobile instead of desktop and tablet/mobile, since the default behavior of the mobile links borders does not show up at the tablet size.

## Testing

1. Visit this PR branch and resize to mobile and back and interact with the menu. The patterns category is the only place there are three levels deep.

## Screenshots

Screen Shot 2020-07-08 at 1 50 13 PM

Screen Shot 2020-07-08 at 1 49 54 PM

Screen Shot 2020-07-08 at 1 49 47 PM


anselmbradford:
There are some layout issues in the mobile menu, but we'll address those in the future as part of a larger mobile menu how it should work task.

docs/_layouts/variation.html
```diff
@@ -1,35 +1,11 @@
...
-                    {% include icons/minus-round.svg %}
-                </span>
-            </span>
-        </button>
-        <div class="o-expandable_content
-                    nav-secondary">
-            {% include sidebar.html %}
-        </div>
-    </aside>
+<main id="main" role="main">
```
contolini, line 4: The GovUK tabs we forked require a parent element to have a `js-enabled` class. We currently add it to the `main` element [here](https://github.com/cfpb/design-system/blob/master/docs/assets/js/main.js#L26). To keep the tabs working, change that selector to `#main` to match this new HTML.

docs/assets/js/sidebar.js
```diff
@@ -0,0 +1,27 @@
+function init() {
+  const secondaryNavCat = document.querySelector( '.ds-nav-container' );
+
+  /**
+   * Test the viewport size and set whether the test passes on the instance.
+   */
+  function handleViewportChange() {
+    if ( window.innerWidth < 601 ) {
+      secondaryNavCat.removeAttribute( 'open' );
```
contolini, line 9: Adding/removing the native `open` attribute is real clever. 👍

docs/assets/js/sidebar.js
```diff
@@ -0,0 +1,27 @@
...
+  // Check viewport state on page load.
+  handleViewportChange();
+
+  // Add event listener for checking viewport state on window resize.
+  window.addEventListener( 'resize', () => {
+    handleViewportChange();
+  } );
+}
+
+export default {
```
contolini, line 25: Trivial: [Other files](https://github.com/cfpb/design-system/blob/master/docs/assets/js/redirect-banner.js#L62) have their export statement on one line.

docs/_layouts/variation.html
anselmbradford: Oops, fixed!

docs/assets/js/sidebar.js
anselmbradford: Thanks!

docs/assets/js/sidebar.js
anselmbradford: Updated!

PR #776, opened 2020-07-08
Rename content guidelines to be guidelines

anselmbradford:
## Changes

- Renames `content_guidelines` section to `guidelines`.

## Testing

1. See PR preview. Pages that said "Content Guidelines" before (see changed files in the PR), should now say "Guidelines"


chosak:
Should [this line](https://github.com/cfpb/design-system/blob/ef52185bacd5494c71025e24973ad4033b0c5c30/docs/_includes/variation-content.html#L233) be changed to make the anchor `id="guidelines"`? Otherwise the jump link from the top is broken, for example [here](https://deploy-preview-776--cfpb-design-system.netlify.app/design-system/foundation/fonts#content-guidelines).

anselmbradford:
> Should this line be changed to make the anchor id="guidelines"? Otherwise the jump link from the top is broken, for example here.

Updated, thanks!

PR #819, opened 2020-07-10
Add toggle all details button

anselmbradford:
## Additions

- Adds a toggle details button that toggles the show/hide button for details.

## Changes

- Makes the edit page button (and the toggle all button) always show text.
 
## Testing

1. Visit the PR preview branch and find a page that has details, such as buttons, and click the toggle details button in the lower-right.

## Screenshots
toggle-details

## Todos

- Is the state of the toggled details retained across pages? If so we need to save a cookie or similar.
- If some details are showing and some others aren't, how should the toggle button work?

### DIFF
--- a/docs/assets/js/main.js
+++ b/docs/assets/js/main.js
@@ -1,6 +1,6 @@
 import {
-  toggleDetails,
-  toggleAllDetails
+  toggleAllDetails,
+  toggleDetails
 } from './toggle-details.js';
 import AnchorJS from 'anchor-js';
 import Expandable from '@cfpb/cfpb-expandables/src/Expandable';
@@ -39,6 +39,18 @@ if ( tabs && tabs.length > 0 ) {
   }
 }
 
+const toggleAllBtn = document.querySelector( '#toggle-details' );
+const toggleBtns = document.querySelectorAll( '.a-toggle_code button' );
+
+toggleAllBtn.addEventListener( 'click', handleToggleAllClick, false );
+
+for ( let i = 0, len = toggleBtns.length; i < len; i++ ) {
+  toggleBtns[i].addEventListener( 'click', handleToggleClick, false );
+}
+
+/**
+ * @param {MouseEvent} event - The mouse event object from the click.
+ */
 function handleToggleAllClick( event ) {
   event.preventDefault();
   toggleAllDetails( toggleAllBtn );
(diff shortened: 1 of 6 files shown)


anselmbradford:
> As a frequent small screen user I prefer the buttons to show text only on hover.

Let's discuss in team time!

> I'd also vote to have the button functionality alternate between Show all details and Hide all details instead of a static Toggle details.

Yes, we discussed this in chat on Friday and I'll update to be Show all details/Hide all details

docs/assets/js/toggle-details.js
```diff
@@ -20,3 +34,34 @@ export function toggleDetails( button, document = window.document ) {
...
+      isShowingAllDetails ? STATE_HIDE : STATE_SHOW
+    );
+  }
+}
+
+export {
+  TOGGLE_ATTRIBUTE,
+  toggleDetails,
+  toggleAllDetails
+}
```
contolini, line 67: Where my newline at

docs/assets/js/toggle-details.js
```diff
@@ -1,16 +1,30 @@
 const HIDDEN_CLASS = 'u-hidden';
-export const TOGGLE_ATTRIBUTE = 'data-toggle-details';
+const TOGGLE_ATTRIBUTE = 'data-toggle-details';
+const STATE_SHOW = 'show';
+const STATE_HIDE = 'hide';
+
+let isShowingAllDetails = true;
```
contolini, line 6: The name of this variable is a little confusing. On page load aren't the details *not* being shown meaning this should start as `false`?

docs/assets/js/main.js
```diff
@@ -36,14 +39,24 @@ if ( tabs && tabs.length > 0 ) {
...
 }
 
-document.addEventListener( 'click', handleDocumentClick, false );
+const toggleAllBtn = document.querySelector( '#toggle-details' );
+const toggleBtns = document.querySelectorAll( '.a-toggle_code button' );
+
+toggleAllBtn.addEventListener( 'click', handleToggleAllClick, false );
+
+for ( let i = 0, len = toggleBtns.length; i < len; i++ ) {
+  toggleBtns[i].addEventListener( 'click', handleToggleClick, false );
```
contolini, line 61: I'm a fan of event delegation by attaching one listener to the document instead of a separate listener for every element but given the simplicity of this website I think either approach is fine.

docs/assets/js/toggle-details.js
anselmbradford: It's true here because I set it equal to its inverse when the toggle function [is called](https://github.com/cfpb/design-system/pull/819/files#diff-3afedd31e202de2df455d7511ae78262R43). I suppose I could move where its set to the end of the function though.

docs/assets/js/main.js
anselmbradford: Yeah I did this because if you have the listener on the document, `event.currentTarget` doesn't work and you have to pick out whether the user clicked the button or the SVG icon it contains.

docs/assets/js/toggle-details.js
anselmbradford: 😱

docs/assets/js/toggle-details.js
anselmbradford: Updated in https://github.com/cfpb/design-system/pull/819/commits/2eaa81e1afe0c35f74677970565af48170a597d1

contolini:
My bad, turns out the browser dimensions had to be enlarged in the webdriver config. I pushed up a commit for ya.

PR #835, opened 2020-07-10
Clean up unused babel and yarn config

anselmbradford:
## Changes

- Remove unused babel and yarn config from /docs directory.
- Include core-js and regenerator-runtime to address https://stackoverflow.com/a/54490329
- Add missing jsdoc

## Testing

1. All should work as before.


docs/admin/src/netlify-cms.js
```diff
@@ -1,3 +1,5 @@
+import "core-js/stable";
+import "regenerator-runtime/runtime";
```
contolini, line 2: Replace these `"` with `'` and this is good-to-go 👍

PR #837, opened 2020-07-13
Add link to CMS instructions page for new users

contolini:
Puts a sentence under pages' title field in the CMS directing users to our how-to page. It's done by creating a new widget that's identical to our string widget with the addition of a hyperlink.

This approach adds some redundant code but adding/editing widgets is the sustainable method of modifying the CMS's left-hand editing pane. The alternative would be forking netlify-cms and tweaking its source code which is a maintenance headache.

## Additions

- `StringWithInstructions` widget

## Testing

1. Everything should function as normal.
1. Try editing a page via the PR preview link below and you should see the new sentence under the title field.

## Screenshots

Screen Shot 2020-07-13 at 12 36 03 PM

## Notes

I wanted to put the help text at the very top of the page, above all the fields BUT netlify doesn't let you touch the area above a field's label. It ended up looking like this which isn't great:

Screen Shot 2020-07-08 at 1 21 53 PM

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/docs/admin/src/widgets/StringWithInstructions.js
+++ b/docs/admin/src/widgets/StringWithInstructions.js
@@ -1,0 +1,64 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const instructionsContainerStyle = {
+  'color': '#444a57',
+  'padding': '20px 0 10px'
+};
+
+const instructionsLinkStyle = {
+  'color': '#444a57',
+  'font-size': '16px',
+  'font-weight': 'normal',
+  'text-decoration': 'underline'
+};
+
+export default class StringTrimmedControl extends React.Component {
+
+  static propTypes = {
+    onChange: PropTypes.func.isRequired,
+    forID: PropTypes.string,
+    value: PropTypes.node,
+    classNameWrapper: PropTypes.string.isRequired,
+    setActiveStyle: PropTypes.func.isRequired,
+    setInactiveStyle: PropTypes.func.isRequired,
+  };
+
+  state = {
+    value: this.props.value || ''
+  };
+
+  handleChange(event) {
+    this.props.onChange( event.target.value.trim() );
+    this.setState( { value: event.target.value } );
+  };
+
+  render() {
+    const {
+      forID,
+      classNameWrapper,
+      setActiveStyle,
+      setInactiveStyle,
+    } = this.props;
+
+    return (
+      <div>
+        <input
+          type='text'
+          id={forID}
+          className={classNameWrapper}
+          value={this.state.value}
+          onChange={event => this.handleChange( event )}
+          onFocus={setActiveStyle}
+          onBlur={setInactiveStyle}
+        />
+        <div style={instructionsContainerStyle}>
+          Need help? Check out our guide on&nbsp;
+          <a href='/design-system/updating-this-website/' target='_blank' style={instructionsLinkStyle}>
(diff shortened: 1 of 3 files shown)


docs/admin/src/widgets/StringWithInstructions.js
```diff
@@ -0,0 +1,64 @@
...
+  'color': '#444a57',
+  'padding': '20px 0 10px'
+};
+
+const instructionsLinkStyle = {
+  'color': '#444a57',
+  'font-size': '16px',
+  'font-weight': 'normal',
+  'text-decoration': 'underline'
+};
```
anselmbradford, line 14: Would it make sense to move this to a stylesheet and import the stylesheet here so that the CSS is in CSS and not a JS object?

docs/admin/src/widgets/StringWithInstructions.js
```diff
@@ -0,0 +1,64 @@
...
+          type='text'
+          id={forID}
+          className={classNameWrapper}
+          value={this.state.value}
+          onChange={event => this.handleChange( event )}
+          onFocus={setActiveStyle}
+          onBlur={setInactiveStyle}
+        />
+        <div style={instructionsContainerStyle}>
+          Need help? Check out our guide on&nbsp;
```
anselmbradford, line 56: Why's the non-breaking space necessary?

docs/admin/src/widgets/StringWithInstructions.js
```diff
@@ -0,0 +1,64 @@
...
+          <a href='/design-system/updating-this-website/' target='_blank' style={instructionsLinkStyle}>
```
anselmbradford, line 57: Add `rel="noopener noreferrer" to this link to avoid the wrath of Lamin. See as an example https://github.com/cfpb/cfgov-refresh/pull/5482

docs/admin/src/widgets/StringWithInstructions.js
contolini: JSX doesn't honor whitespace so there are a [few ways](https://stackoverflow.com/questions/40264084/best-practice-when-adding-whitespace-in-jsx) to force them. I *could* just put the `<a>` element on the same line as L56 but then it'd be super long and fugly so I settled on `&nbsp;`.

docs/admin/src/widgets/StringWithInstructions.js
contolini: This is such a one-off thing that I was avoiding that BUT we do have a stylesheet just for netlify so that seems like a great place for it. 👍

docs/admin/src/widgets/StringWithInstructions.js
Scotchester: That is so depressing.

docs/admin/src/widgets/StringWithInstructions.js
contolini: It turns out it's easy to add stylesheets to the preview pane but there's no method of adding one to the editing pane. :/

PR #920, opened 2020-07-17
Fix filterable table example

anselmbradford:
## Changes

- Add missing closing div for filter.
- Remove unused `o-expandable__filters` class.
- Remove erroneously added `data-bound=true` attribute that should be added by the JS.

## Testing

1. Visit tables page in PR preview and go to the filterable table example and the "filter" should be able to be opened.

## Screenshots

Screen Shot 2020-07-17 at 10 48 35 AM

### DIFF
--- a/docs/pages/tables.md
+++ b/docs/pages/tables.md
@@ -555,8 +555,8 @@ variation_groups:
           * Please note the importance of defining a `thead` and `tbody` to preserve the table’s header through sorting operations.
       - variation_name: Filtering tables
         variation_code_snippet: >
-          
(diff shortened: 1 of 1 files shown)


docs/pages/tables.md
```diff
@@ -575,6 +575,7 @@ variation_groups:
               <div class="o-expandable_content o-expandable_content__transition o-expandable_content__collapsed" aria-expanded="false" style="max-height: 0px;">
                   <p>Filter options go here</p>
               </div>
+          </div>
```
contolini, line 578: Oof.

PR #949, opened 2020-07-20
Fix iOS navigation bug

anselmbradford:
iOS triggers a resize event when scrolling that was causing the navigation menu to close unexpectedly. This PR first checks if the window size has actually changed before closing the menu.

## Changes

- Check that window size has changed before closing or opening the sidebar navigation menu.

## Testing

1. Pull branch and run locally. 
2. Open the iOS simulator (Applications > Xcode > [right-click Show Package Contents] > Contents > Developer > Applications > Simulator.app)
3. Open the localhost:4000/design-system in the Simulator.
4. Open the navigation menu.
5. Scroll down the page. The menu should not close on its own.

### DIFF
--- a/docs/assets/js/sidebar.js
+++ b/docs/assets/js/sidebar.js
@@ -5,15 +5,26 @@
 function init() {
   const secondaryNavCat = document.querySelector( '.ds-nav-container' );
 
+  // First collapse the navigation if in mobile.
+  if ( innerWidth < 601 ) {
+    secondaryNavCat.removeAttribute( 'open' );
+  }
+
+  let windowWidth = window.innerWidth;
   /**
    * Test the viewport size and set whether the test passes on the instance.
    */
   function handleViewportChange() {
-    if ( window.innerWidth < 601 ) {
-      secondaryNavCat.removeAttribute( 'open' );
-
-    } else {
-      secondaryNavCat.setAttribute( 'open', 'open' );
+    // Collapse the navigation if we resize to mobile,
+    // but only if we haven't already.
+    // Otherwise, we're on desktop size, so open the navigation.
+    let innerWidth = window.innerWidth;
+    if ( innerWidth !== windowWidth ) {
+      if ( innerWidth < 601 ) {
+        secondaryNavCat.removeAttribute( 'open' );
+      } else {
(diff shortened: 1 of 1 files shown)


docs/assets/js/sidebar.js
```diff
@@ -5,15 +5,26 @@
 function init() {
   const secondaryNavCat = document.querySelector( '.ds-nav-container' );
 
+  // First collapse the navigation if in mobile.
+  if ( innerWidth < 601 ) {
```
contolini, line 9: For clarity maybe make this `window.innerWidth` to emphasize you're checking the global value.

docs/assets/js/sidebar.js
```diff
@@ -5,15 +5,26 @@
...
-    if ( window.innerWidth < 601 ) {
-      secondaryNavCat.removeAttribute( 'open' );
-
-    } else {
-      secondaryNavCat.setAttribute( 'open', 'open' );
+    // Collapse the navigation if we resize to mobile,
+    // but only if we haven't already.
+    // Otherwise, we're on desktop size, so open the navigation.
+    let innerWidth = window.innerWidth;
+    if ( innerWidth !== windowWidth ) {
```
contolini, line 22: I'd vote for either returning early:

```
if ( innerWidth === windowWidth ) {
  return;
}
```

or combining the conditionals:

```
if ( innerWidth !== windowWidth && innerWidth < 601 ) {
  secondaryNavCat.removeAttribute( 'open' );
} else {
  secondaryNavCat.setAttribute( 'open', 'open' );
}
```

docs/assets/js/sidebar.js
contolini: Or actually, couldn't you move `let windowWidth = window.innerWidth;` above this line and use `windowWidth` here?

docs/assets/js/sidebar.js
anselmbradford: I went with the first, as the second version would run the `else` in cases where the original structure wouldn't. https://github.com/cfpb/design-system/pull/949/commits/e8d367df2ea004033e59632c3ff4d8c8d91bb9cf

PR #1030, opened 2020-07-29
Run docs CSS through webpack

anselmbradford:
Development build = unminified and includes sourcemaps.
Production build = minified and includes ~~no~~ sourcemaps.

## Changes

- Create clearer delineated dev and prod webpack configuration for doc site assets.
- Send doc site CSS through webpack less loader and process it with postcss and autoprefixer.
- Remove stand-alone `watch` commands from package.json and include them in the `start` command.
- Lint configuration files.
- Add extra directory step for reset files path in cfpb-core and cfpb-grid.

## Testing

1. Pull branch and run `yarn && yarn start`
2. Visit http://localhost:4000/design-system/components/dropdowns-and-multi-selects and inspect the drop-down `select` element. It should have `-webkit-appearance: none;`. Compare to the live site, which should only have `appearance: none;` (and it'll be crossed-out).

**Testing sourcemaps:**
1. Run `yarn run compile-dev` and check `docs/dist/css/` and see that there's an unminified `main.css` and a sourcemap.
2. Run `yarn run compile-prod` and check `docs/dist/css/ and see that there's a minified `main.css` and a sourcemap.

## Notes

- When do we want to make a development build? When do we want to make a production build? Do we want the option to switch between them locally through a command-line command?

## Todo

- This just handles the docs site build process, but the packages' build process should also be aligned with these steps.


contolini:
> When do we want to make a development build? When do we want to make a production build? Do we want the option to switch between them locally through a command-line command?

What if we *always* do a development build (with the exception of the `/dist/js/netlify-cms.js` file). Due to the nature of this website, I imagine visitors will be frequently inspecting the source code and it'd be nice to make that experience as pleasant as possible by not obfuscating any assets. The downside is the performance hit. Are the unminified assets significantly larger than the minified ones?

anselmbradford:
> What if we always do a development build (with the exception of the /dist/js/netlify-cms.js file). Due to the nature of this website, I imagine visitors will be frequently inspecting the source code and it'd be nice to make that experience as pleasant as possible by not obfuscating any assets. The downside is the performance hit. Are the unminified assets significantly larger than the minified ones?

I don't really like selectively having some files in the dev build and some in the prod build. The optimizer shaves off about 33%. You can run `yarn run compile-dev` and `yarn run compile-prod` to see the difference in `docs/dist/`

We coooould include the sourcemap in the production build, which I think would address your concern, if it works correctly on the minified files.

contolini:
> We coooould include the sourcemap in the production build, which I think would address your concern, if it works correctly on the minified files.

Oooh that's a great compromise. Sourcemaps are only downloaded after dev tools opens so if they're separate files and not inlined, there's no performance hit.

anselmbradford:
@contolini This is ready for review again with the update to add sourcemaps.

PR #1043, opened 2020-08-10
Don't run browser tests on forks

chosak:
The browser tests require Sauce credentials to be defined in GitHub repo settings (SAUCE_USERNAME and SAUCE_ACCESS_KEY). If these credentials aren't defined, the tests will fail. Unfortunately this means that if you have a fork without a Sauce account, you can't pass the test.

This change modifies the browser tests Action so that it won't run when on a fork. It would be better to somehow disable the tests only if those credentials aren't defined, but I couldn't find an easy way to do so.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -59,6 +59,11 @@ jobs:
         sleep 5
 
     - name: Run browser tests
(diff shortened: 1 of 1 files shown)


chosak:
@anselmbradford thanks for the review. Yes, I believe so.

PR #1050, opened 2020-08-27
Expand expandables when printing

contolini:
Expand expandables when printing so that users get the entire page's content.

See https://github.local/CFPB/el-camino/issues/299

## Changes

- Hide the netlify CMS floating buttons when printing because why not. They block the page text.
- Hide the `Show`/`Hide` buttons when printing because they're irrelevant (and if we left them, it'd incorrectly say `Show` above expandables that are now open when printed)
- Sets a large max-height when printing because we display the raw URLs next to links which causes the expandable's content height to be greater than the non-print height, causing this bug:

3a0a0300-e6c2-11ea-8345-8a0365b5283a

## Testing

1. Visit the PR preview link below and view a print preview of the expandables page. They should all be expanded.

## Screenshots

Screen Shot 2020-08-27 at 12 55 45 PM

## Notes

-

## Todos

-

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/docs/assets/css/netlify.less
+++ b/docs/assets/css/netlify.less
@@ -32,9 +32,9 @@
 	}
 
 	// Hide the admin buttons when printing
-	.respond-to-print({
+	.respond-to-print( {
 		display: none;
-	});
+	} );
 }
 
 // Don't show the toggle code button in the Netlify CMS preview pane
--- a/docs/assets/css/netlify.less
+++ b/docs/assets/css/netlify.less
@@ -30,6 +30,11 @@
 			margin-left: 0.4375em;
 		}
 	}
+
+	// Hide the admin buttons when printing
+	.respond-to-print({
+		display: none;
+	});
 }
 
 // Don't show the toggle code button in the Netlify CMS preview pane
--- a/packages/cfpb-expandables/src/cfpb-expandables.less
+++ b/packages/cfpb-expandables/src/cfpb-expandables.less
@@ -62,14 +62,12 @@
             display: none;
         }
 
-        &__expanded .o-expandable_cue-close,
+        &__expanded .o-expandable_cue-close {
+            display: block;
+        }
+
         &__collapsed .o-expandable_cue-open {
             display: block;
-
-            // Hide the interactive expandable cues when printing
-            .respond-to-print({
-                display: none;
-            });
         }
     }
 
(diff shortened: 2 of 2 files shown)


docs/assets/css/netlify.less
```diff
@@ -30,6 +30,11 @@
 			margin-left: 0.4375em;
 		}
 	}
+
+	// Hide the admin buttons when printing
+	.respond-to-print({
```
anselmbradford, line 35: Very minor, but in our other "respond-to" utilities, we include a space inside the parens https://github.com/cfpb/design-system/blob/f4613641062df12454f3101e09232a7b94096f0e/packages/cfpb-layout/src/organisms/featured-content-module.less#L9

anselmbradford:
Would hiding the "Show Details" link on print be in scope here?

contolini:
> Would hiding the "Show Details" link on print be in scope here?

Normally no but I threw it in because without it, it displays the wrong verb in the printed content (`Show` when it's already shown due to new print CSS rule):

Screen Shot 2020-08-27 at 3 08 08 PM

That's not a huge deal though and I'm happy to remove it if it's overreaching.

packages/cfpb-expandables/src/cfpb-expandables.less
```diff
@@ -84,6 +86,14 @@
 
         &__collapsed {
             max-height: 0;
+
+            // Ensure all expandables are expanded when printing.
+            // To accommodate print stylesheets that display the raw URL after links,
+            // set an enormous max height to accommodate expandables that have a lot of links.
+            .respond-to-print({
+                display: block;
+                max-height: 99999px !important;
```
anselmbradford, line 95: You could also consider aggregating all the print styles at the bottom and have the utility only once:

```
.respond-to-print( {
  .o-expandable_target__expanded .o-expandable_cue-close,
  .o-expandable_target__collapsed .o-expandable_cue-open {
    display: none;
  }
  
  .o-expandable_content__collapsed {
    display: block;
    max-height: 99999px;
  }
} );
```

The advantage is you only need to work on one area of the file if working on print styles. Also, I _think_ the `!important` wouldn't be needed. The disadvantage is if the associated styles are changed, they could be overlooked in the print styles.

contolini: I like the idea of moving them to the bottom. The `!important` is required to override the inline `max-height` style set via [JS](https://github.com/cfpb/design-system/blob/master/packages/cfpb-expandables/src/ExpandableTransition.js#L100).

anselmbradford: Oooo right :(

PR #1062, opened 2020-09-11
Fix plugin path to remove hardcoding

anselmbradford:
Plugin positon in the less plugins index will vary by the environment this file is built in, so we need to scan the plugins instead of having a hardcoded position.

## Changes

- Make plugin index position dynamic.

## Testing

1. `yarn build` should pass.


contolini (approved):
DO IT

PR #1066, opened 2020-09-30
Update to autoprefixer 10.x

anselmbradford:
## Changes

- Update to autoprefixer 10.x
- Adds postcss as a peer dependency (see [changelog](https://github.com/postcss/autoprefixer/blob/master/CHANGELOG.md#100-alis-volat-propriis))

## Testing

1. PR branch should build


contolini (approved):
Buonissimo! Sauce is complaining about an IE11 thing that we're working on.

PR #1080, opened 2020-10-06
cfpb-core: Indent numbered lists

anselmbradford:
## Changes

- Indents numbered lists 0.3rem to avoid clipping that can happen when lists are double digits. 

## Testing

1. Check lists page in the design system on this PR preview link.

## Screenshots

Screen Shot 2020-10-06 at 12 06 23 PM

Before:

Screen Shot 2020-10-06 at 12 03 56 PM

After:

Screen Shot 2020-10-06 at 12 04 19 PM

### DIFF
--- a/docs/pages/lists.md
+++ b/docs/pages/lists.md
@@ -55,6 +55,17 @@ variation_groups:
           presented in the order of importance, or a finite list of a counted
           number of items.
         variation_code_snippet: |-
+          <p>Less than 10 items is not indented.</p>
+          <ol>
+              <li>List item 1</li>
+              <li>List item 2</li>
+              <li>List item 3</li>
+          </ol>
+
+          <p>
+            More than 9 items is indented to align double digits.
+            Inner lists are not indented, regardless of number of items.
+          </p>
           <ol>
               <li>List item 1</li>
               <li>
@@ -70,9 +81,16 @@ variation_groups:
                         <li>Nested item 2-8</li>
                         <li>Nested item 2-9</li>
                         <li>Nested item 2-10</li>
+
                   </ol>
               </li>
-              <li>List item 3</li>
+              <li>List item 3
+                  <ol>
+                        <li>Nested item 3-1</li>
+                        <li>Nested item 3-2</li>
+                        <li>Nested item 3-3</li>
+                  </ol>
+              </li>
               <li>List item 4</li>
               <li>List item 5</li>
               <li>List item 6</li>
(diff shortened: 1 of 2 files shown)


niqjohnson (commented):
Sorry, I'm totally jumping in here, so you may have already talked through these questions as you were working (in which case, feel free to ignore!). But, a few questions that came to mind:

1. Did designers weigh in on this change? I vaguely remember that having list items flush left was an intentional design decision back in the day, though clearly we didn't account for how that would affect two-digit counters. But this change makes one-digit counters ever so slightly indented, which is both a change from the original style and feels—to me, anyway—in that zone of large enough to be noticeable but still small enough to look unintentional and misaligned, especially for lists that have only one-digit counters. There might not be much we can do to avoid that, but I wanted to see what other designers thought.
2. If we do want to go with this 0.3 rem extra left margin, any reason we wouldn't want to add it to `base.less` lines 460–464, where it looks like we're already trying to do something similar to keep counters out of the margin?
https://github.com/cfpb/design-system/blob/33a68476ffce763f1523d269b27d517151fd61b0/packages/cfpb-core/src/base.less#L460-L464

niqjohnson:
@anselmbradford: I was chatting with @huetingj and @nataliafitzgerald about this in our working group today, and I had another idea. It sounds like you and @huetingj already went over some options and settled on the 0.3 rem nudge for ordered list items as the best option, but @nataliafitzgerald made the excellent point that it seems a shame to nudge *all* ordered lists out of flush left alignment just to account for the handful of long lists we have (and that we discourage content authors from making, anyway). Which got me thinking…

What if we add the 0.3 rem nudge to *only* ordered lists that contain 10 or more items? I think [the `:nth-last-child()` method in this A List Apart article](https://alistapart.com/article/quantity-queries-for-css/#section6) might do the trick, right?

anselmbradford:
Hmmmmmm. interesting idea regarding applying it only to the longer lists. My initial hunch is using nth-child may be difficult to apply to the items before item 10, but I'll try it out. Also, I think you're right this can be integrated into the existing padding. I'll circle back to this. Thanks!

stephanieosan:
Could we merge this in for now and consider adding the padding only when lists are 10+ items as a separate feature? 

I think having digits visibly cut off (which is the current state of things across the site) is worse than the slightly different alignment.

anselmbradford:
@niqjohnson This has been updated to only indent lists greater than 10 items long:

Screen Shot 2020-10-16 at 12 04 44 PM

niqjohnson (approved):
Very cool, that CSS quantity query method is my new favorite thing! This is looking good to me on the lists page and seems like it'll fix the cut-off digits we see for long lists on cf.gov.

anselmbradford:
@niqjohnson All credit to you for digging it up! I just entered it in.

stephanieosan:
disco-cat

PR #1094, opened 2020-11-05
Migrate to ES6 export/import syntax

anselmbradford:
## Changes

- Migrate package JavaScript to more modern ES6 export/import syntax.

## Testing

1. PR tests should pass and PR preview should look okay and error-free.


test/unit-test/src/cfpb-atomic-component/src/components/AtomicComponent-spec.js
```diff
@@ -1,5 +1,4 @@
-const srcPath = require( '../src-path' );
-let AtomicComponent = require( srcPath + '/components/AtomicComponent' );
+import AtomicComponent from '../../../../../../packages/cfpb-atomic-component/src/components/AtomicComponent';
```
contolini, line 1: What I visualize when I see relative paths like this:

elevator

It's a spec file though so no biggie.

contolini (approved):
Everything builds a-okay and looks good. There might be downstream issues if Babel isn't configured correctly? Something to keep an eye on.

PR #1095, opened 2020-11-05
Remove extra atomic component layers

anselmbradford:
We define each atomic layer, Atom, Molecule, Organism, Page, Template, as a JS structure that defines its hierarchy. However, we don't do anything with this hierarchy and it obfuscates what uses the base AtomicComponent.

## Removals

- Remove extra atomic JS structure files.

## Testing

1. PR preview should work.


contolini (approved):
Made sense at the time but no more. 👍

PR #1098, opened 2020-11-12
Remove object shorthand to aid IE11

anselmbradford:
IE11 does not like ES6 object shorthand. Instead of polyfilling all the places this is used, this PR just removes the shorthand.

## Changes

- Remove ES6 object shorthand to support IE11 limitations.

## Testing

1. PR tests should pass.


contolini (approved):
wfm 👍

PR #1100, opened 2020-11-16
Add datepicker markup

anselmbradford:
## Additions

- Adds date picker inputs to the DS.

## Testing

1. See the text inputs page on the PR preview.

## Screenshots

Screen Shot 2020-11-16 at 11 31 36 AM

## Notes

This has the bug where the input increases in width when first mousing over it. Let me know if you see a reason that is happening. It appears to be DS specific, since it doesn't happen in the blog filter in cf.gov.

### DIFF
--- a/docs/pages/text-inputs.md
+++ b/docs/pages/text-inputs.md
@@ -120,6 +120,17 @@ variation_groups:
               <textarea class="a-text-input a-text-input__full"
                         id="full-textarea-example">Placeholder text</textarea>
           </div>
+      - variation_name: Datepicker inputs
+        variation_code_snippet: >-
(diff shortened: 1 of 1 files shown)


contolini (approved):
Removing our custom font prevents the hover bug. The datepicker's shadow DOM might be basing the width of the container on Arial, a split second before Avenir Next is downloaded and rendered, and hovering triggers a repaint. 🤷 

datepicker

The offending element is `::-webkit-datetime-edit`:

element

anselmbradford:
Hmmmmm good snooping. I'll merge this to get the markup in and we can look closer if it's an issue.

PR #1102, opened 2020-11-17
Update normalize paths

anselmbradford:
The current paths to the normalize CSS works in the cf.gov repo, but breaks here. I think if we remove the path it'll just look into the `node_modules` where ever it's used.

## Changes

- Update normalize paths

## Testing

1. `gulp build` should pass.


contolini (approved):
Builds great 👍

PR #1110, opened 2020-11-23
Fix linter error and warnings

anselmbradford:
## Changes

- Fix linter error and warnings

## Testing

1. PR tests should pass.


contolini (approved):
Locally it looks like there are a handful of eslint complexity complaints but that's a much bigger fish to fry.

PR #1112, opened 2020-11-23
Change data-bound to data-js-hook value

anselmbradford:
We have two syntaxes for specifying that a component is initialized. `data-bound` and `data-js-hook`. This PR changes the value to `data-js-hook` used elsewhere.

## Changes

- Change data-bound to data-js-hook value

## Testing

1. PR checks should pass.


contolini (approved):
wfm 👍

PR #1114, opened 2020-11-25
Fix coverage generation

anselmbradford:
Coverage was committed to the repo and shouldn't be. Also, it was not being generated locally.

## Changes

- Fixes Jest CLI coverage flags to allow generation of code coverage locally.
- Removes ignored coverage files from the repo.

## Testing

1. Pull branch and run `gulp test`
2. Coverage should be generated at test/unit-test-coverage/lcov-report/index.html (which can be opened in a web browser).


contolini (approved):
🔪

PR #1118, opened 2020-12-01
Remove utility individual packages from cfpb-atomic-component

anselmbradford:
## Changes

- Remove utility individual packages from cfpb-atomic-component.
- Move utility readmes to main cfpb-atomic-component readme.

## Testing

1. All should be as it were. E.g. PR checks should pass.


contolini (approved):
TY

PR #1119, opened 2020-12-01
Fix multiselect filtering

contolini:
While exploring the IE11 issue I discovered the cause is a larger issue with the multiselect not correctly filtering options.

## Changes

- Pass the indices of the filtered options to the `_filterMatches` function.

## Testing

1. Visit https://cfpb.github.io/design-system/components/dropdowns-and-multiselects and enter some text like "super" into the multiselect field. It should only display option 8 that contains the text "super". Instead, it shows everything but option 8.
1. Visit the PR preview link and verify it works as expected.

### DIFF
--- a/packages/cfpb-forms/src/organisms/Multiselect.js
+++ b/packages/cfpb-forms/src/organisms/Multiselect.js
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines-per-function */
 // Required modules.
 import EventObserver from '../../../cfpb-atomic-component/src/mixins/EventObserver.js';
 import MultiselectModel from './MultiselectModel.js';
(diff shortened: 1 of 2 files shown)


packages/cfpb-forms/src/organisms/Multiselect.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines-per-function */
```
anselmbradford, line 1: What was this added for? If it was for `function Multiselect( element ) {` we should only disable it for that line, not the whole file. We'd add `// eslint-disable-line max-statements max-lines-per-function` to line 21.

anselmbradford (changes requested):
This was my mistake of incorrectly refactoring two forEach calls to a regular for loops for compatibility with IE in https://github.com/cfpb/design-system/pull/1096. I really mucked them up 😭 

The idea of MultiselectModel is to make the multiselect a little more MVC, so all data processing happens in the model and the main Multiselect file just queries the model for the state of things. So with that in mind, what do you think about these changes to fix this?

Screen Shot 2020-12-02 at 11 20 13 AM

contolini:
@anselmbradford Yeah as long as `filterIndices` is getting defined, that works for me! Do you still have those changes locally? Want to just push to this branch?

contolini:
Cool, checks are passing and it's working great 👍

anselmbradford (approved):

PR #1120, opened 2020-12-04
Multiselect: refactor atomic component configuration to allow consistent init flag config

anselmbradford:
## Additions

- `atomic-helpers.js` utilities from cf.gov.
- `data-hook.js` from cf.gov
- `standard-type.js` from cf.gov.

## Removals

- Removes `config.js` and merges it into `standard-type.js`

## Changes

- Requires a call to `init()` method after an AtomicComponent instance is created.
- Moves AtomicComponent tag name to a constant.
- Incorporates atomic-helper utilities `instantiateAll` and `setInitFlag` into `AtomicComponent`.
- Modifies `instantiateAll` method from cf.gov version to allow passing of a `scope` argument.
- Change imported package paths to not be relative and instead to look in the `@cfpb` package.

## Testing

1. Check the PR preview and see that the multiselect has a `data-js-hook="state_atomic_init"` flag set.
2. Expandables and Tables functionality should be the same as before.


packages/cfpb-atomic-component/src/utilities/data-hook.js
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * @param {HTMLNode} element - DOM element.
+ * @param {string} value
+ *   Value to add to the element's JS data-* hook.
+ * @returns {string} The value that was added.
+ * @throws {Error} If supplied value contains a space,
+ *   meaning it would be two values, which is likely a typo.
+ */
+function add( element, value ) {
+  if ( value.indexOf( ' ' ) !== -1 ) {
```
contolini, line 13: Is this a common mistake we make? Why choose to check for spaces but not other illegal characters?

contolini (approved):
Great implementation, works as expected and I dig the concise functions. 👍

packages/cfpb-atomic-component/src/utilities/data-hook.js
anselmbradford: It's because when we run `contains` it chunks the value in the data- attribute by spaces. The check here is not so much whether an individual value is wrong, but more that it only create one wrong value instead of two if it is (since a space would later on produce two attribute values). https://github.com/cfpb/design-system/pull/1120/files/b5ea4866e8c3474fff1f37ba648a25b0ccb3c383#diff-b7cc11f19fbb7b8ead3d107f312f63a5521ce0c08e15d7d593d04b9ae25c459fR57

 I suppose we could check if the value itself is wrong as well with a regex of some sort huh.

contolini: Hmmm on second thought I'd rather not police what people are using as their `data` hook values. If they're inserting weird characters they should handle those errors on their own.

PR #1121, opened 2020-12-09
Ensure correct multiselect input labels and IDs

chosak:
This change properly handles multiselects defined with an `id` but not a `name`, as in [the example in the design system docs](https://cfpb.github.io/design-system/components/dropdowns-and-multiselects#multiselects-1).

```html
<select class="o-multiselect" id="something" multiple>
```

Currently, the JS code generates a text `<input>` with an empty `id` attribute. This means that the associated `<label>` points to the wrong thing, causing failures in accessibility audits, [for example](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1607472905780-28501.report.html).

Additionally, the current behavior fails to ensure that the generated checkboxes have unique `id`s, simply copying the option `value`s. This commit prepends each `id` with the enclosing `input` ID, ensuring uniqueness on the page.

### DIFF
--- a/packages/cfpb-forms/src/organisms/Multiselect.js
+++ b/packages/cfpb-forms/src/organisms/Multiselect.js
@@ -71,7 +71,7 @@ function Multiselect( element ) { // eslint-disable-line max-statements
     }
 
     _instance = this;
-    _name = _dom.name;
+    _name = _dom.name || _dom.id;
     _placeholder = _dom.getAttribute( 'placeholder' );
     _options = _dom.options || [];
 
@@ -172,8 +172,10 @@ function Multiselect( element ) { // eslint-disable-line max-statements
         'class': 'm-form-field m-form-field__checkbox'
       } );
 
+      const _checkboxId = _name + '-' + option.value;
+
       MultiselectUtils.create( 'input', {
-        'id':     option.value,
+        'id':     _checkboxId,
         // Type must come before value or IE fails
         'type':    'checkbox',
         'value':   option.value,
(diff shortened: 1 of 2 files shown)


packages/cfpb-forms/src/organisms/Multiselect.js
```diff
@@ -172,8 +172,10 @@ function Multiselect( element ) { // eslint-disable-line max-statements
         'class': 'm-form-field m-form-field__checkbox'
       } );
 
+      const _checkboxId = _name + '-' + option.value;
```
anselmbradford, line 175: Maybe we should process `option.value` here and ensure it doesn't contain any spaces. So:

`const checkboxId = _name + '-' + option.value.trim().replace( /\s+/g, '-' ).toLowerCase();`

anselmbradford: You don't need the leading underscore on `checkboxId`. The leading underscore is for variables defined outside a method that are used in the method, but aren't exported publicly. We're not consistent with this though :-\ as I see `_optionsItemDom` defined above has the underscore, while things like `selectionsItemDom` do not.

test/unit-test/src/cfpb-forms/src/organisms/Multiselect-spec.js
```diff
@@ -79,7 +79,7 @@ describe( 'Multiselect', () => {
       search.click();
       simulateEvent( 'keydown', search, { keyCode: 40 } );
 
-      expect( document.activeElement.id ).toBe( 'Debt collection' );
+      expect( document.activeElement.value ).toBe( 'Debt collection' );
```
anselmbradford, line 82: Maybe we should also check the id. I think this should be correct based on the review changes:

`expect( document.activeElement.id ).toBe( 'test-select-debt-collection' );`

anselmbradford (changes requested):

anselmbradford (approved):

PR #1125, opened 2020-12-16
Multiselect: Remove list classes mashup

anselmbradford:
We have markup in the multiselect that looks like `<ul class="m-list m-list_unstyled o-multiselect_options">…`. Applying two components to the same dom node is vulnerable to outside issues bleeding into the component. This PR removes the list classes and applies the one style rule they're needed for directly to the multiselect element.

## Removals

- List classes from multiselect component.

## Changes

- Applies left padding removal directly to `o-multiselect_choices`.

## Testing

1. Compare this PR's preview multiselect to that on https://cfpb.github.io/design-system/components/dropdowns-and-multiselects and make sure the selected items and drop-down have the same padding.


packages/cfpb-forms/src/organisms/multiselect.less
```diff
@@ -92,6 +92,7 @@ select.o-multiselect {
     }
 
     &_choices {
+        padding-left: 0;
```
contolini, line 95: Wow, that's all that needed to be transferred? Easy peasy.

contolini (approved):

PR #1126, opened 2020-12-16
Fix multiselect removal bug

contolini:
The multiselect option IDs weren't being transferred to the removal buttons making it impossible to remove a selected option.

## Changes

- Moves the `checkboxId` value to a property of `option` so that we can use it when rendering the selected options.

## Testing

1. You should be able to remove selected multiselect options from this PR preview while you currently cannot at https://cfpb.github.io/design-system/components/dropdowns-and-multiselects
1. `yarn test:browser` should pass.


contolini:
Looks like IE11 is complaining. I shall investigattte.

PR #1127, opened 2020-12-16
Encode inline SVGs for IE11 support

anselmbradford:
## Changes

- URL encodes inline SVGs for IE11 support.

## Testing

1. Visit the PR preview in IE11 in Sauce Labs and see that the inline background image SVGs are showing up on the dropdowns and in the checkboxes in the multiselect.

## Screenshots

Screen Shot 2020-12-16 at 3 31 13 PM
Screen Shot 2020-12-16 at 3 31 18 PM


contolini (approved):
They're back!

PR #1128, opened 2020-12-16
Multiselect: Pass clicks of the selected item's button down to the label it contains

anselmbradford:
## Changes

- Listen for clicks of the selected item's button and pass that click down to the label it contains.

## Testing

1. Visit the PR preview in IE11 in Sauce Labs and see that the multiselect's selected items can be removed.


contolini (approved):
And just that like, we have passing tests again 🎉

PR #1130, opened 2020-12-16
Multiselect: Explicitly set button type

anselmbradford:
If a button appears in a form it'll be assumed by default that it is the submit button. We don't want this to happen when we drop a multiselect into a form, so we need to explicitly set the `type` to be `button`. 

## Additions

- Multiselect: Add `type="button"` attribute to selected item buttons.

## Testing

1. Visit multiselect in PR preview and see that selected items can be added and removed, and when inspected have `type="button"`.


contolini (approved):
TIL

PR #1131, opened 2020-12-17
Update dom-traverse.js

anselmbradford:
Design System had older version of dom-traverse utility.

## Changes

- Updates to latest dom-traverse utility.

## Testing

1. PR checks should pass.

### DIFF
--- a/packages/cfpb-atomic-component/src/utilities/dom-traverse.js
+++ b/packages/cfpb-atomic-component/src/utilities/dom-traverse.js
@@ -13,11 +13,15 @@ function queryOne( expr, con ) {
 }
 
 /**
- * Get the nearest parent node of an element.
+ * Traverse the element and its parents (heading toward the document root)
+ * until a node is found that matches the provided selector string.
+ * Will return itself or the matching ancestor.
+ * If no such element exists, it returns null.
  *
  * @param {HTMLNode} elem - A DOM element.
  * @param {string} selector - CSS selector.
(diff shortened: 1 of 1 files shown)


packages/cfpb-atomic-component/src/utilities/dom-traverse.js
```diff
@@ -15,34 +15,47 @@ function queryOne( expr, con ) {
...
+ * @param {HTMLNode} elem - A DOM element.
  * @param {string} selector - CSS selector.
  * @returns {HTMLNode} Nearest parent node that matches the selector.
  */
-function closest( element, selector ) {
-  if ( 'closest' in element ) {
-    return element.closest( selector );
-  }
+function closest( elem, selector ) {
+  elem = elem.parentNode;
```
contolini, line 23: This line is causing the tests to fail. We immediately override elem with its parent which makes sense because we're looking for the "nearest parent node of an element" but it's a departure from the current functionality.

packages/cfpb-atomic-component/src/utilities/dom-traverse.js
```diff
@@ -15,34 +15,53 @@ function queryOne( expr, con ) {
 /**
  * Get the nearest parent node of an element.
```
contolini, line 16: Can you mention in the JSDoc comment that traversal begins with the element and it'll return itself if applicable?

contolini (approved):

PR #1134, opened 2020-12-30
Update transitions and add transitions page to DS

anselmbradford:
Transitions add an event listener that listens for the `transitionEnd` event, however, the element may have transitions that are performed outside of the transition class, making their `transitionEnd` event fire early and multiple times. To get around this, I've added a `CSS_PROPERTY` value that gets passed into the `BaseTransition` constructor, that is then checked in the `transitionEnd` handler when it fires.

## Additions

- Adds transitions page and example code to design system.

## Changes

- Adds `CSS_PROPERTY` value to transitions to pass to `BaseTransition` and associated logic.
- Updates transition tests to match those in cf.gov repo.

## Testing

1. Check patterns > transition patterns in the PR's preview.

## Screenshots

Screen Shot 2020-12-30 at 3 58 57 PM

### DIFF
--- a/docs/assets/js/main.js
+++ b/docs/assets/js/main.js
@@ -5,6 +5,8 @@ import {
 import AnchorJS from 'anchor-js';
 import Expandable from '@cfpb/cfpb-expandables/src/Expandable';
 import Multiselect from '@cfpb/cfpb-forms/src/organisms/Multiselect';
+import AlphaTransition from '@cfpb/cfpb-atomic-component/src/utilities/transition/AlphaTransition.js';
+import MoveTransition from '@cfpb/cfpb-atomic-component/src/utilities/transition/MoveTransition.js';
 import Table from '@cfpb/cfpb-tables/src/Table';
 import { Tabs } from 'govuk-frontend';
 import redirectBanner from './redirect-banner.js';
@@ -34,6 +36,23 @@ if ( multiselectDom ) {
 
 Expandable.init();
 Table.init();
+
+// Transition example code.
+const moveTransitionExample = document.querySelector( '.example-box.u-move-transition' );
+if ( moveTransitionExample !== null ) {
+  const moveTransition = new MoveTransition( moveTransitionExample ).init();
+  moveTransitionExample.addEventListener( 'click', () => {
+    moveTransition.moveRight();
+  } );
+}
+
+const alphaTransitionExample = document.querySelector( '.example-box.u-alpha-transition' );
+if ( alphaTransitionExample !== null ) {
+  const alphaTransition = new AlphaTransition( alphaTransitionExample ).init();
+  alphaTransitionExample.addEventListener( 'click', () => {
+    alphaTransition.fadeOut();
+  } );
+}
 
 const main = document.querySelector( '#main' );
 const tabs = document.querySelectorAll( '[data-module="tabs"]' );
--- a/docs/assets/js/main.js
+++ b/docs/assets/js/main.js
@@ -51,6 +51,9 @@ if ( alphaTransitionExample !== null ) {
   const alphaTransition = new AlphaTransition( alphaTransitionExample ).init();
   alphaTransitionExample.addEventListener( 'click', () => {
     alphaTransition.fadeOut();
+    setTimeout( () => {
+      alphaTransition.fadeIn();
+    }, 2000 );
   } );
 }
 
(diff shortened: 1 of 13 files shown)


docs/assets/js/main.js
```diff
@@ -35,6 +37,23 @@ if ( multiselectDom ) {
...
+  moveTransitionExample.addEventListener( 'click', () => {
+    moveTransition.moveRight();
+  } );
+}
+
+const alphaTransitionExample = document.querySelector( '.example-box.u-alpha-transition' );
+if ( alphaTransitionExample !== null ) {
+  const alphaTransition = new AlphaTransition( alphaTransitionExample ).init();
+  alphaTransitionExample.addEventListener( 'click', () => {
+    alphaTransition.fadeOut();
```
contolini, line 53: To make it easier for devs to inspect the element and repeatedly click it, can we have the example fade back in?

```suggestion
    alphaTransition.fadeOut();
    setTimeout( () => {
      alphaTransition.fadeIn();
    }, 2000 );
```

anselmbradford: I'll add that for the move one too.

packages/cfpb-atomic-component/src/utilities/transition/AlphaTransition.js
```diff
@@ -1,12 +1,13 @@
...
 import BaseTransition from './BaseTransition.js';
-import EventObserver from '../../mixins/EventObserver.js';
+import EventObserver from '@cfpb/cfpb-atomic-component/src/mixins/EventObserver.js';
 
 // Exported constants.
 const CLASSES = {
-  BASE_CLASS: 'u-alpha-transition',
-  ALPHA_100:  'u-alpha-100',
-  ALPHA_0:    'u-alpha-0'
+  CSS_PROPERTY: 'opacity',
```
anselmbradford, line 7: @contolini 

This addition is a bummer because it has to match the CSS that gets transitioned in `u-alpha-transition`. I discovered that the fade is not working on the feedback form on https://www.consumerfinance.gov/ask-cfpb/what-is-a-debt-to-income-ratio-why-is-the-43-debt-to-income-ratio-important-en-1791/ and the reason is the `background-color` is being transitioned there for some reason, which fires the `transitionEnd` on the target element before the fade has been performed. The change in this PR of passing in the CSS property that is being transitioned was my solution for this, which is then checked in the `_transitionComplete` handler. Let me know if you see a better way.

docs/pages/transition-patterns.md
```diff
@@ -0,0 +1,79 @@
...
+section: patterns
+status: Released
+description: Transition patterns are animations that happen when a user
+  interacts with an element on the page. They are CSS transition styles that are
+  controlled via JavaScript.
+variation_groups:
+  - variations:
+      - variation_name: Move transition
+        variation_description: A transition that moves from one position to another.
+        variation_code_snippet: <div class="u-move-transition example-box">Click me!</div>
```
contolini, line 13: Is there a way to keep the example JS next to the example HTML? What if we changed the code snippet to:

```html
<div class="u-move-transition example-box">Click me!</div>
<script>
    const moveTransitionExample = document.querySelector( '.example-box.u-move-transition' );
    const moveTransition = new MoveTransition( moveTransitionExample ).init();
    moveTransitionExample.addEventListener( 'click', () => {
        moveTransition.moveRight();
    } );
</script>
```

And then removed the JS from `main.js`? IMHO it would make the implementation details clearer.

anselmbradford: Hmmmmmmmmmm yea that makes sense.

docs/pages/transition-patterns.md
```diff
@@ -0,0 +1,131 @@
...
+          transition.init();
+
+
+          // Call methods on the transition.
+
+          transition.moveRight();
+
+          ```
+      - variation_name: Alpha transition
+        variation_description: A transition that fades from one position to another.
```
contolini, line 68: Should this read

```suggestion
        variation_description: A transition that fades from one opacity to another.
```

?

anselmbradford: Oops

anselmbradford: Yes your update is correct. Copy pasted from MoveTransition.

packages/cfpb-atomic-component/src/utilities/transition/AlphaTransition.js
contolini: That makes sense. FWIW it is nice knowing which property the transition modifies without having to hunt down the applicable CSS file.

contolini (approved):

PR #1140, opened 2021-01-07
Transitions: Call transitionComplete handler if transition duration is zero

anselmbradford:
`transitionEnd` event will not fire if the transition duration is zero, which leads to the `u-is-animating` class from not being removed. We therefore should manually called the transitionComplete event handler immediately if the `u-no-animation` class is present.

## Changes

- When using a transition, call the transitionComplete event handler immediately if the `u-no-animation` class is present.
- Add test to check for proper removal of `u-is-animating` class if transition duration is zero.

## Testing

1. PR checks should pass.


contolini (approved):

contolini:
Why, oh why, did netlify spam this pr...

PR #1152, opened 2021-01-12
Update "data-visualization-1" page

ielerol:
Automatically generated by Netlify CMS


ielerol:
I created a new version of the "titles" image so it's just an example without any written specs, and did some restructuring of the lists of variations. I'd appreciate some eyes on the changes to see if others agree it's an improvement. @huetingj @nataliafitzgerald what do you think of [the preview?](https://deploy-preview-1152--cfpb-design-system.netlify.app/design-system/guidelines/data-visualization-guidelines#chart-components)

huetingj:
Hi! I think this is a great improvement. The specs were so difficult to read (and inaccessible) in the previous version, so I love simply labeling the example and then stating the specs in the page’s text itself. In the future, we should apply this treatment to the Labels and keys section. 

Seeing this improvement made me remember how much restructuring this page needs in general, oiy—the fact that all sections are sitting within the treatment meant only to house a pattern’s different types; that the introductory text is labeled “Introduction” unlike any other DS page; etc. I think some restructuring could possibly be done to this "Chart components" section as well. It throws me a bit that the "Titles" visual example is not directly followed by the listed specs. I wondered at first if the specs should be nested in a “Show details” expandable. But I’m not sure. It’d be nice to look at the page’s structure holistically before rearranging. Hopefully we’ll be able to pick that up soon! Thanks for this improvement in the meantime. :)

ielerol:
Oh yeah, more room for improvement here for sure, I just wanted to concentrate on a piece that could be tackled in a 2-hour time block. Thank you!

Issue #1052, opened 2020-09-01
Table pattern: Update row header style


caheberer:
**Is your feature request related to a problem? Please describe.**
All caps headers work well for 1 or 2 word labels, but most of our content does not fit this convention and is often longer, making the labels difficult to read. 

**Describe the solution you'd like**
We're proposing changing table headers from H5 to paragraph medium, sentence case, and top-aligning the baseline.

**New specs:**
Background: Gray-05 #F8F8F8
Font: Avenir Next Medium, sentence case
Font-color: Black #101820
Font-size: 16px
Top-align baseline

**Acceptance criteria**
- [ ] Design has reviewed and 👍 
- [ ] Specs updated on design system page
- [ ] Code snippet added to design system page

anselmbradford:
PR --> https://github.com/cfpb/design-system/pull/1154

PR #1154, opened 2021-01-12
Tables: Adjust table header style

anselmbradford:
## Changes

- Changes table headers from H5 to paragraph medium, sentence case, and top-aligning the baseline.

## Testing

1. Check tables page in this PR and compare table headers to https://cfpb.github.io/design-system/components/tables

## Screenshots

Before:
Screen Shot 2021-01-12 at 1 21 24 PM

After:
Screen Shot 2021-01-12 at 1 30 35 PM

### DIFF
--- a/packages/cfpb-core/src/base.less
+++ b/packages/cfpb-core/src/base.less
@@ -485,11 +485,12 @@ td {
     padding: unit( 10px / @base-font-size-px, em );
 
     thead & {
-        .h5();
-
-        padding: unit( 10px / @font-size, em );
+        // 10px / 14px
+        padding: unit( 10px / @size-v, em );
         background: @table-head-bg;
         color: @table-head-text;
+        font-size: unit( 16px / @base-font-size-px, em );
+        vertical-align: top;
     }
 }
 


sonnakim:
Looks good to me @anselmbradford. So nice to say goodbye to all caps for the table headers. Only thing I notice is that [the request](https://github.com/cfpb/design-system/issues/1052) asks for a background color of #F8F8F8, which looks like it would be a change from the #f7f8f9 we usually use (https://www.consumerfinance.gov/consumer-tools/educator-tools/resources-for-older-adults/financial-security-as-you-age/),  I'm not sure if they're specifically asking to change the background color or if maybe they had the wrong spec listed?

anselmbradford:
>  I'm not sure if they're specifically asking to change the background color or if maybe they had the wrong spec listed?

Yea I asked this in chat on makeathon day and my understanding is the f8f8f8 is incorrect in the issue.

huetingj:
Yah, #F8F8F8 isn't in [our palette](https://cfpb.github.io/design-system/foundation/color). But I just noticed that we state #F8F8F8 in the design specs on the [Tables page](https://cfpb.github.io/design-system/components/tables). Aiy aiy. 

Keep it coded as #f7f8f9. I'll update the Tables DS page now to state #f7f8f9.

And I like the look of paragraph medium, sentence case, top-aligning the baseline!

sonnakim:
Good catch @huetingj ! Thanks for making that update.

contolini (approved):
🚀

PR #1159, opened 2021-01-19
Add unit test for data-hook and instantiateAll

anselmbradford:
## Changes

- Add unit test spec for `data-hook.js`.
- Adjust data-hook import to import specific methods.
- Add additional `instantiateAll` test.

## Testing

1. `gulp test:unit` should pass.

### DIFF
--- a/packages/cfpb-atomic-component/src/utilities/atomic-helpers.js
+++ b/packages/cfpb-atomic-component/src/utilities/atomic-helpers.js
@@ -9,7 +9,7 @@
    - Atom
    ========================================================================= */
 
-import * as dataHook from './data-hook';
+import { add, contains, remove } from './data-hook';
 import { STATE_PREFIX } from './standard-type';
 
 /**
@@ -86,11 +86,11 @@ function _verifyClassExists( element, baseClass ) {
  *   false otherwise.
  */
 function setInitFlag( element ) {
-  if ( dataHook.contains( element, INIT_FLAG ) ) {
+  if ( contains( element, INIT_FLAG ) ) {
     return false;
   }
 
-  dataHook.add( element, INIT_FLAG );
+  add( element, INIT_FLAG );
 
   return true;
 }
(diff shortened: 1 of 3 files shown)


test/unit-test/src/cfpb-atomic-component/src/utilities/data-hook-spec.js
```diff
@@ -0,0 +1,47 @@
...
+  } );
+
+  describe( 'remove()', () => {
+    it( 'should remove a value to the data-* attribute of the element', () => {
+      remove( testComponent, 'test_behavior' );
+      expect( testComponent.getAttribute( 'data-js-hook' ) ).toBe( '' );
+    } );
+  } );
+
+  describe( 'remove()', () => {
```
contolini, line 41: The `remove()` test is in here twice.

packages/cfpb-atomic-component/src/utilities/atomic-helpers.js
```diff
@@ -9,7 +9,7 @@
    - Atom
    ========================================================================= */
 
-import * as dataHook from './data-hook';
+import { add, contains, remove } from './data-hook';
```
contolini, line 12: Why import the individual methods separately? For clarity?

contolini (approved):

test/unit-test/src/cfpb-atomic-component/src/utilities/data-hook-spec.js
anselmbradford: 🤔 not sure how that happened, but fixed!

packages/cfpb-atomic-component/src/utilities/atomic-helpers.js
anselmbradford: Yeah so if you search for the methods you will get the import statement in the search results. Also, if data-hook has more added to it, it won't be automatically pulled into this file where it's unneeded.

PR #1167, opened 2021-01-25
Multiselect: Only initialize if we're not on a mobile OS

anselmbradford:
The multiselect should degrade to the native multiple select element when we're on mobile, so that the native multiple select UI is used. This PR adds a utility to check the user agent string for a mobile OS and bails out of initializing the multiselect if it is present.

## Changes

- Adds media utility for checking user agent string for mobile OSs.
- Bails out of initializing the multiselect if the user agent is a mobile OS.
- Merges unused `a-multiselect` into `o-multiselect` CSS class.

## Testing

1. Test the PR preview on mobile devices in sauce labs and visit the multiselect and see that it is degraded and the device uses the native multiple select UI.

## Screenshots

Some mobile examples:
Screen Shot 2021-01-25 at 10 35 34 AM

Screen Shot 2021-01-25 at 10 36 03 AM

Screen Shot 2021-01-25 at 10 38 16 AM

Screen Shot 2021-01-25 at 10 46 51 AM


contolini (approved):
Interestingly, iPad Pro reports itself as a desktop (`MacIntel`) which shows the fragility of user agent sniffing but this solution is a great improvement for mobile users. 👍

anselmbradford:
> Interestingly, iPad Pro reports itself as a desktop (MacIntel) which shows the fragility of user agent sniffing but this solution is a great improvement for mobile users. 👍

I did notice that, but it seemed a big enough screen area that maybe we'd want the desktop experience there anyway, so I left it.

PR #1175, opened 2021-01-28
Move imports to top and changes to absolute imports

anselmbradford:
`gulp build` was not passing because the `transitions.less` file needs a variable from `vars.less`. `vars.less` needs a variable from `brand-colors.less`.

## Changes

- Move imports to top of file.
- Changes to absolute imports.
- Add reference import to `brand-colors.less` inside `vars.less`.

## Testing

1. `gulp build` should pass.
2. DS PR preview should be unchanged from production.


contolini (approved):
Builds real good 👍

PR #1177, opened 2021-01-29
Multiselect: Remove redundant _optionsData array

anselmbradford:
`_optionsData` array is in `MultiselectModel` and `Multiselect`. It should only be in `MultiselectModel`.

## Changes

- Remove `_optionsData` array from `Multiselect`.
- Includes `data-index` attribute from https://github.com/cfpb/design-system/pull/1176

## Testing

1. Check dropdowns and multiselects page on the preview PR. The multiselect should be able to add and remove options same as before.

### DIFF
--- a/packages/cfpb-forms/src/organisms/Multiselect.js
+++ b/packages/cfpb-forms/src/organisms/Multiselect.js
@@ -243,7 +243,10 @@ function Multiselect( element ) { // eslint-disable-line max-statements
    * @returns {string} A hopefully unique ID.
    */
   function _getOptionId( option ) {
-    return _name + '-' + option.value.trim().replace( /\s+/g, '-' ).toLowerCase();
+    /* Replace any character that is not a word character with a dash.
+       https://regex101.com/r/ShHmRw/1
+    */
+    return _name + '-' + option.value.trim().replace( /[^\w]/g, '-' ).toLowerCase();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


packages/cfpb-forms/src/organisms/Multiselect.js
```diff
@@ -233,6 +237,15 @@ function Multiselect( element ) { // eslint-disable-line max-statements
     selectionsItemLabelDom.addEventListener( 'keydown', _selectionKeyDownHandler );
   }
 
+  /**
+   * Create a unique ID based on a select's option HTML element.
+   * @param {HTMLNode} option - A option HTML element.
+   * @returns {string} A hopefully unique ID.
```
contolini, line 243: 😆

packages/cfpb-forms/src/organisms/Multiselect.js
```diff
@@ -233,6 +237,15 @@ function Multiselect( element ) { // eslint-disable-line max-statements
...
+   */
+  function _getOptionId( option ) {
+    return _name + '-' + option.value.trim().replace( /\s+/g, '-' ).toLowerCase();
```
contolini, line 246: Instead of only replacing spaces should we be replacing all non-word characters? https://stackoverflow.com/a/4374890

contolini (approved):

packages/cfpb-forms/src/organisms/Multiselect.js
anselmbradford: Kudos to that developer/regex for posting a still relevant solution 10 years later.

Updated in https://github.com/cfpb/design-system/pull/1177/commits/b888465c69841a6a8686737ef549adc24f534b11

PR #1178, opened 2021-02-01
Expand multiselect option targets

contolini:
Remove all non-clickable area from around multiselect options to improve the UX and prevent a bug that causes misclicked multiselects to get permanently stuck open.

I was originally drafting a JS solution to the bug but a CSS solution ended up being much cleaner.

See https://github.local/CFPB/el-camino/issues/366

## Changes

- Add padding around the individual options instead of their container.

## Testing

1. Visit the PR preview link and confirm the problem described in the above issue no longer exists.

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/packages/cfpb-forms/src/organisms/multiselect.less
+++ b/packages/cfpb-forms/src/organisms/multiselect.less
@@ -93,12 +93,10 @@ select.o-multiselect {
 
     &_options {
         background-color: @white;
-        padding: unit( 10px / @base-font-size-px, em );
-        // Double padding on the right to accommodate the fieldset's forced y scrollbar above
-        padding-right: unit( 20px / @base-font-size-px, em );
+        padding: 0;
 
-        li:last-child {
-            margin-bottom: 0;
+        li {
+            margin: 0;
         }
 
         &.u-filtered li:not(.u-filter-match) {
@@ -125,6 +123,9 @@ select.o-multiselect {
         }
 
         .a-label {
+            box-sizing: border-box;
+            padding-top: unit( 10px / @base-font-size-px, em );
+            padding-left: unit( 10px / @base-font-size-px, em );
             width: 100%;
         }
     }


anselmbradford:
I can still reproduce the bug by clicking on the edge of the drop-down. Sorry :(

Also, are we sure it's not an issue that the edge of the toggles comes right up to the bottom of the checkboxes? It seems like that might be a violation of the clickable target area on mobile:

Screen Shot 2021-02-01 at 12 05 17 PM

contolini:
> I can still reproduce the bug by clicking on the edge of the drop-down. Sorry :(

Curse you edge!

> Also, are we sure it's not an issue that the edge of the toggles comes right up to the bottom of the checkboxes? It seems like that might be a violation of the clickable target area on mobile:
> 
> Screen Shot 2021-02-01 at 12 05 17 PM

Yeah the padding should prob be split between top and bottom. Let me rethink this solution...

contolini:
@anselmbradford I added a check to verify that the user clicked on an option and not a nondescript area of the fieldset. I also did a better job vertically spacing the options.

packages/cfpb-forms/src/organisms/multiselect.less
```diff
@@ -125,6 +129,11 @@ select.o-multiselect {
         }
 
         .a-label {
+            box-sizing: border-box;
+            padding: unit( 5px / @base-font-size-px, em )
```
anselmbradford, line 133: Since this is taking up four lines anyway, let's maybe spell out `padding-top`, `padding-right`, etc. and drop the shorthand so we don't have to remember what order the shorthand goes in.

anselmbradford (approved):

PR #1179, opened 2021-02-03
Makes a-label__heading respond like an h4 on small screens

wpears:
`.a-label__heading` is designed to function as an `h4` (via its call to the `.heading-4` mixin) but doesn't respond to small screens [like an `h4`](https://github.com/cfpb/design-system/blob/main/packages/cfpb-core/src/base.less#L250-L256). This PR fixes that.

### DIFF
--- a/packages/cfpb-forms/src/atoms/label.less
+++ b/packages/cfpb-forms/src/atoms/label.less
@@ -30,5 +30,13 @@
                 margin-top: unit( 10px / @base-font-size-px, em );
             }
         }
+
+        .respond-to-max( @bp-xs-max, {
+            @h4-font-size-on-xs: @base-font-size-px;
+
+            margin-bottom: unit( 10px / @h4-font-size-on-xs, em );
+            font-size: unit( @h4-font-size-on-xs / @base-font-size-px, em );
+            line-height: unit( 18px / @h4-font-size-on-xs );
+        } );
     }
 }
--- a/packages/cfpb-forms/src/atoms/label.less
+++ b/packages/cfpb-forms/src/atoms/label.less
@@ -14,7 +14,7 @@
     }
 
     &__heading {
-        .heading-4();
+        .h4();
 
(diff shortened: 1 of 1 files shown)


packages/cfpb-forms/src/atoms/label.less
```diff
@@ -30,5 +30,13 @@
                 margin-top: unit( 10px / @base-font-size-px, em );
             }
         }
+
+        .respond-to-max( @bp-xs-max, {
+            @h4-font-size-on-xs: @base-font-size-px;
```
anselmbradford, line 35: Why is defining a new variable `@h4-font-size-on-xs` necessary?

wpears: The other definition is scoped the the `.respond-to-max`, so this one needs to be as well. I could pull the `@h4-font-size-on-xs` definition out (or just use `base-font-size-px` directly) instead. Thoughts?

wpears: I think the original reason for the defining the variable was to make a point that, instead of stepping down one heading level (like h1-h4) we're just defining a different font size.

anselmbradford: Are you talking about the variable defined [here](https://github.com/cfpb/design-system/blob/d67f9b6af2ee320ef0baed1e97e904c38c8ea9a9/packages/cfpb-core/src/base.less#L65)? I was thinking just use `@base-font-size-px` directly. Seems like a question for @Scotchester

Scotchester: How about just changing the mixin that `.a-label__heading` includes to `.h4()` and then it will inherit the media query?

wpears: Aha! Updated in d678fe4a

Scotchester (approved):

PR #1180, opened 2021-02-03
Multiselect: Make arrow open/close drop-down

anselmbradford:
When you open the multiselect, let's make it so the down-arrow on the right-hand side closes it when clicked, yaaay! Also, let's flip the arrow when it is open! Yaaaah!

## Changes

- Make clicking the arrow in the search input open AND close the drop-down.
- Switch the drop-down arrow to an up arrow when the drop-down is open.

## Testing

1. Open the PR preview and go to the dropdowns and multiselects page and open and close the multiselect by clicking the arrow.

## Screenshots

mm-dropdown

### DIFF
--- a/packages/cfpb-forms/src/organisms/Multiselect.js
+++ b/packages/cfpb-forms/src/organisms/Multiselect.js
@@ -124,7 +124,7 @@ function Multiselect( element ) { // eslint-disable-line max-statements
     _fieldsetDom.classList.add( 'u-invisible' );
     _fieldsetDom.setAttribute( 'aria-hidden', true );
     _model.resetIndex();
-    _instance.dispatchEvent( 'expandEnd', { target: _instance } );
+    _instance.dispatchEvent( 'collapseBegin', { target: _instance } );
 
     return _instance;
   }
@@ -413,6 +413,20 @@ function Multiselect( element ) { // eslint-disable-line max-statements
         target.style.cursor = 'pointer';
       } else {
         target.style.cursor = 'auto';
+      }
+    } );
+
+    _headerDom.addEventListener( 'mouseup', function( event ) {
+      const target = event.target;
+
+      /* Check if we're over the down-arrow on the right side of the input.
+         Also check if the fieldset is open.
+         35 = width of the arrow on the right of the search input.
+         140 = the max-height value set in multiselect.less for the fieldset.
+      */
+      if ( event.layerX > target.offsetWidth - 35 &&
+           _fieldsetDom.offsetHeight === 140 ) {
+        _searchDom.blur();
       }
     } );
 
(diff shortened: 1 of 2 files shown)


huetingj:
looks good!

packages/cfpb-forms/src/organisms/Multiselect.js
```diff
@@ -416,6 +417,20 @@ function Multiselect( element ) { // eslint-disable-line max-statements
...
+    _headerDom.addEventListener( 'mouseup', function( event ) {
+      const target = event.target;
+
+      /* Check if we're over the down-arrow on the right side of the input.
+         Also check if the fieldset is open.
+         35 = width of the arrow on the right of the search input.
+         140 = the max-height value set in multiselect.less for the fieldset.
+      */
+      if ( event.layerX > target.offsetWidth - 35 &&
+           _fieldsetDom.offsetHeight === 140 ) {
```
Scotchester, line 429: Is this the best way we have to check that the fieldset is open?

Scotchester (approved):
Works as advertised, but I do have one question.

packages/cfpb-forms/src/organisms/Multiselect.js
```diff
@@ -416,6 +417,20 @@ function Multiselect( element ) { // eslint-disable-line max-statements
...
 
+    _headerDom.addEventListener( 'mouseup', function( event ) {
+      const target = event.target;
+
+      /* Check if we're over the down-arrow on the right side of the input.
+         Also check if the fieldset is open.
+         35 = width of the arrow on the right of the search input.
+         140 = the max-height value set in multiselect.less for the fieldset.
+      */
+      if ( event.layerX > target.offsetWidth - 35 &&
```
contolini, line 428: Should we be using [offsetX](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/offsetX) instead of [layerX](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/layerX)? They operate the same but the former is in a spec while the latter is deprecated.

packages/cfpb-forms/src/organisms/Multiselect.js
anselmbradford: Yeah I know it's hackery :( I think we need to do some re-writing of the drop-down. It opens when the search input gets focus and closes when it's blurred. Focus happens on mousedown, so it gets the attributes of being open when it is clicked, not when the drop-down has actually completed opening. Getting the height was the only way I found to reliably find that it was actually open. We should re-write this to dispatch events for the life cycle of the drop-down expand/collapse and then listen to the `expandEnd` event (or actually probably the `transitionEnd` event) and add attributes at that point, which this if statement would then query. As noted in this PR, the collapse method is what dispatches (erroneously I think) the `expandEnd` event.

packages/cfpb-forms/src/organisms/Multiselect.js
anselmbradford: Yes, good catch. It would be nice to find a better way to target the dropdown in the future…

packages/cfpb-forms/src/organisms/Multiselect.js
Scotchester: Makes sense. This is fine for now, but could you capture everything you've said here in a new issue? (Or do we already have one open?)

anselmbradford: https://github.com/cfpb/design-system/issues/1182

PR #1184, opened 2021-02-04
Bugfix: multiselect missing padding on max selection

anselmbradford:
## Changes

- Adds padding to max selections view of drop-down.
- Adds `pointer-events: none` to max selections view, so that clicking the drop-down closes it.

## Testing

1. In PR preview open multiselects page and interact with multiselect to select five options. Max selections view should have padding and close when clicked.

## Screenshots

Screen Shot 2021-02-04 at 11 48 33 AM

## Todos

- Arrow doesn't work to close the drop-down (on max selection view) as we don't have a good way yet to know when the drop-down is completely open.

### DIFF
--- a/packages/cfpb-forms/src/organisms/multiselect.less
+++ b/packages/cfpb-forms/src/organisms/multiselect.less
@@ -131,8 +131,12 @@ select.o-multiselect {
             content: 'No results found';
         }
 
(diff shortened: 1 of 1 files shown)


contolini (approved):
Things are weird. But now they're slightly less weird.

PR #1187, opened 2021-02-05
MaxHeightTransition: Convert summary transition to px from ems

anselmbradford:
Probably best to show a fixed height for the summary, regardless on base font size. This changes the max-height to pixels from ems, and match the value set in the Summary component https://github.com/cfpb/consumerfinance.gov/blob/c1bc097d598ffae6f1c2215e1bcac3c057f972d1/cfgov/unprocessed/js/organisms/Summary.js#L45

## Changes

- Change `u-max-height-summary` `max-height` to pixels from ems.
- Set the `max-height` on the element when the page has fully loaded, as the reported value from `scrollHeight` may be incorrect before then.

## Testing

1. max-height transition on the /design-system/patterns/transition-patterns page in the PR preview should work as before.
2. Change the font-size in the Chrome settings and reload the page. See that the box is the same height, whereas on https://cfpb.github.io/design-system/patterns/transition-patterns it is taller.
3. Expand the box by clicking it and the content should not be cropped at the bottom, whereas on https://cfpb.github.io/design-system/patterns/transition-patterns it may be (because the max-height was calculated before the page fully loaded).

## Screenshots

Before:
Screen Shot 2021-02-05 at 3 23 46 PM

After:
Screen Shot 2021-02-05 at 3 25 02 PM

### DIFF
--- a/packages/cfpb-atomic-component/src/utilities/transition/MaxHeightTransition.js
+++ b/packages/cfpb-atomic-component/src/utilities/transition/MaxHeightTransition.js
@@ -31,7 +31,7 @@ function MaxHeightTransition( element ) {
   function init() {
     _baseTransition.init();
 
-    /* The scrollHeight of an element may incorrect if the page hasn't
+    /* The scrollHeight of an element may be incorrect if the page hasn't
        fully loaded yet, so we listen for that to happen before calculating
        the element max-height. */
     window.addEventListener( 'load', _pageLoaded );
--- a/packages/cfpb-atomic-component/src/utilities/transition/MaxHeightTransition.js
+++ b/packages/cfpb-atomic-component/src/utilities/transition/MaxHeightTransition.js
@@ -31,7 +31,10 @@ function MaxHeightTransition( element ) {
   function init() {
     _baseTransition.init();
 
-    refresh();
+    /* The scrollHeight of an element may incorrect if the page hasn't
+       fully loaded yet, so we listen for that to happen before calculating
+       the element max-height. */
+    window.addEventListener( 'load', _pageLoaded );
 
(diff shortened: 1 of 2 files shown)


packages/cfpb-atomic-component/src/utilities/transition/MaxHeightTransition.js
```diff
@@ -31,7 +31,10 @@ function MaxHeightTransition( element ) {
   function init() {
     _baseTransition.init();
 
-    refresh();
+    /* The scrollHeight of an element may incorrect if the page hasn't
```
contolini, line 34: may *be* incorrect

packages/cfpb-atomic-component/src/utilities/transition/MaxHeightTransition.js
```diff
@@ -42,6 +45,15 @@ function MaxHeightTransition( element ) {
     return this;
   }
 
+  /**
+   * The whole page has loaded,
+   * including all dependent resources such as stylesheets and images.
+   */
+  function _pageLoaded() {
+    window.removeEventListener( 'load', _pageLoaded );
```
contolini, line 53: This would be a good use case for the [`once` parameter](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#parameters) but it looks like IE doesn't support it. :(

contolini (approved):
🚀 I don't see any cropping on prod but it's likely because of the lightning quick page load.

packages/cfpb-atomic-component/src/utilities/transition/MaxHeightTransition.js
anselmbradford: TIL

PR #1196, opened 2021-02-17
Multiselect: Change search input ID to DOM ID

anselmbradford:
If the multiselect's underlying select element has a `name` attribute set, the search input field `id` becomes that name. However, on pages such as https://www.consumerfinance.gov/about-us/blog/, the Topic filter `for` attribute is expected to point to the `id`, not the `name`. 

## Changes

- Sets the search input to be the `id` value of the underlying `select` element, instead of the `name` attribute, if present.

## Testing

1. Check the PR preview and visit the multiselect page and see that you can click the label above the multiselect and see that it activates the search input.


contolini (approved):
Nice catch 👍

PR #1203, opened 2021-03-01
Change mobile jump link background to @gray-5

### LINKED ISSUE #1201: Links: Jump links with background modifier have insufficient contrast
**Describe the bug**

[Jump links with background modifier](https://cfpb.github.io/design-system/components/links#jump-link-with-background-modifier) (grey background on mobile) fail Lighthouse checks with "Background and foreground colors do not have a sufficient contrast ratio."

This was previously mentioned at https://github.com/cfpb/design-system/issues/902#issuecomment-661332012:

> The documented pattern for jump links with a grey background fail accessibility contrast check, according to [the WebAIM contrast checker](https://webaim.org/resources/contrastchecker/). `a-link__bg` uses `jump-link_bg` [which is defined as`@gray-10`](https://github.com/cfpb/design-system/blob/f4613641062df12454f3101e09232a7b94096f0e/packages/cfpb-typography/src/cfpb-typography.less#L54), which is #e7e8e9, and the link color is #0072ce, which gives a contrast ratio of 3.98:1. The threshold for "normal text" is 4.5:1.
> 
> @huetingj suggests we should be using `@gray-5` instead (#f7f8f9), which is what is being used [on the Bureau structure page](https://www.consumerfinance.gov/about-us/the-bureau/bureau-structure/). This passes the contrast check with a ratio of 4.59:1, and it turns out this is happening due to custom CSS in cfgov-refresh that sets overriding styles on the BS elements.
>
> In chat we discussed a path forward of modifying `a-link__bg` to instead use `@gray-5`, and then removing the customization from the BS page. @anselmbradford does this seem reasonable?

Since that time we've redesigned the BS page to no longer use this pattern. A quick check of the nightly crawl shows that we have no pages on cf.gov right now that use `a-link__bg`.

This error causes failures in the nightly Lighthouse accessibility checks we run on the design system, and has since #1188 a few weeks back. I'm not sure why this was passing previously if this has been an issue for a while. That PR bumped the version of Lighthouse we use to [its 7.0 release](https://github.com/GoogleChrome/lighthouse/releases/tag/v7.0.0), which includes an upgrade of the underlying axe-core library from 3.5.5 to 4.1.1. A quick look at the [axe-core release notes](https://github.com/dequelabs/axe-core/releases) shows a few contrast-related changes, but I'm not sure which one could be changing how Lighthouse treats this case.

**To Reproduce**

In Chrome, run a Lighthouse Accessibility audit using the Mobile device. You will see the error "Background and foreground colors do not have a sufficient contrast ratio." for the `span.a-link_text` in `<a class="a-link__bg">`.

**Screenshots**
image

chosak:
This improves the accessibility of jump links on mobile by increasing the contrast between the foreground text and background color.

Fixes #1201.

## Screenshots

image

## Checklist

- [x] Any _change_ in functionality is tested
- [x] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [x] Project documentation has been updated
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### DIFF
--- a/docs/pages/variables.md
+++ b/docs/pages/variables.md
@@ -358,7 +358,7 @@ variation_groups:
           @link-underline-active:  @navy;
 
           // .a-link__jump
-          @jump-link_bg:           @gray-10;
+          @jump-link_bg:           @gray-5;
           @jump-link_border:       @gray-40;
 
           // code
--- a/packages/cfpb-typography/src/cfpb-typography.less
+++ b/packages/cfpb-typography/src/cfpb-typography.less
@@ -55,7 +55,7 @@
 // Links
 
 // .a-link__jump
-@jump-link_bg: @gray-10;
+@jump-link_bg: @gray-5;
 @jump-link_border: @gray-40;
 
 //
--- a/packages/cfpb-typography/usage.md
+++ b/packages/cfpb-typography/usage.md
@@ -80,7 +80,7 @@ Color variables referenced in comments are from [@cfpb/cfpb-core's brand-colors.
 // Links
 
 // .a-link__jump
-@jump-link_bg:              @gray-10;
+@jump-link_bg:              @gray-5;
 @jump-link_border:          @gray-40;
 ```
 


anselmbradford:
We need it updated in the docs too:

- https://github.com/cfpb/design-system/blame/main/docs/pages/variables.md#L361
- https://github.com/cfpb/design-system/blame/main/packages/cfpb-typography/usage.md#L83

Scotchester (approved):

huetingj:
The new background color looks great. Thanks @chosak!

Scotchester (approved):

PR #1207, opened 2021-03-02
Add variation group name to fix missing h2

chosak:
The carousel anti-pattern page currently fails Lighthouse accessibility testing because it lacks an h2:

"Heading elements are not in a sequentially-descending order"

This causes the nightly Lighthouse tests to fail, for example:

https://github.com/cfpb/design-system/actions/runs/612274827

### DIFF
--- a/docs/pages/carousel.md
+++ b/docs/pages/carousel.md
@@ -15,7 +15,7 @@ description: >-
 
   Carousels offer poor usability and are considered an [anti-pattern](https://thegood.com/insights/ecommerce-image-carousels/). Their use is discouraged.
 variation_groups:
-  - variation_group_name: ""
+  - variation_group_name: Types
     variations:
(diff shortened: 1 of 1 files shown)


chosak:
@contolini FYI the nightly Lighthouse checks run against [every page in the DS](https://github.com/cfpb/design-system/blob/76edfb93504ff8cba8f058fc5b43641c2038f9e2/lighthouserc.js#L15-L19), whether it is linked from navigation or not.

With this change, the carousel page now gets a "Types" heading:

image

This bumps the Lighthouse a11y score to 100%, and can be tested on the preview branch above: https://deploy-preview-1207--cfpb-design-system.netlify.app/design-system/patterns/carousel

contolini (approved):
Lighthouse FTW

PR #1215, opened 2021-03-26
Bugfix: Add gulp build to release step

anselmbradford:
When releasing if you don't manually run `gulp build` it won't run with the release. Meaning the released compiled JS may not be up-to-date.

## Changes

- Bugfix: Add gulp build to release step.
- Re-generate DS JS file.

## Testing

1. PR checks should pass.


contolini (approved):
Wow nice catch 👍

PR #1211, opened 2021-03-10
Updated outdated dependencies: core-js, marked , netlify-cms, slugify, babel/core, babel/plugin-proposal-class-properties, babel/plugin-proposal-object-rest-spread, babel/preset-env, autoprefixer, chromedriver, concurrently, css-loader, html-loader, lerna, mini-css-extract-plugin, postcss, postcss-loader, stylelint, stylelint-config-standard, svgo, webpack

anselmbradford:
## Changes

- Updated outdated dependencies: core-js, marked , netlify-cms, slugify, babel/core, babel/plugin-proposal-class-properties, babel/plugin-proposal-object-rest-spread, babel/preset-env, autoprefixer, chromedriver, concurrently, css-loader, html-loader, lerna, mini-css-extract-plugin, postcss, postcss-loader, stylelint, stylelint-config-standard, svgo, webpack

## Testing

1. PR checks should pass.

PR #1216, opened 2021-03-29
SVGO config fixes

anselmbradford:
## Changes

- Fix linting errors in `svgo.config.js` file.
- Re-process icons.

## Testing

1. `yarn run process-icon-svgs` should pass, but not change any files.


contolini (approved):
Looks like these new SVGs are coming from the updated version of svgo in https://github.com/cfpb/design-system/pull/1211

Issue #1239, opened 2021-06-03
Markdown tables aren't rendering properly


sonnakim:
**Describe the bug**
Markdown tables are getting squished, so that columns are super narrow and images are too small to see:
![image](https://user-images.githubusercontent.com/9451445/120675070-bfe18880-c462-11eb-9917-c24b01a8f7c7.png)

This is especially bad for the pages section, which uses images in tables to show differences in page layouts. Example on the [landing page](https://cfpb.github.io/design-system/pages/landing-pages).  

**Expected behavior**
Unsquished tables

**Notes**
Browsers: This bug appears on Chrome 91 but not Chrome 90, not the latest Safari.

contolini:
Note: This issue surfaced with Chrome >= 91 when its [table width calculations](https://chromium-review.googlesource.com/c/chromium/src/+/2875946) were fixed.

sonnakim:
🙌  thanks @contolini !

sonnakim:
The types markdown tables look wonderfully roomy and de-squished now  👏  but it seems markdown tables in other sections of the page are still in need of unsquishing: see e.g. [this markdown table in a behavior section](https://cfpb.github.io/design-system/pages/landing-pages#behavior), [another markdown table in a behavior section](https://cfpb.github.io/design-system/patterns/featured-content-module#behavior), and [this table in a guidelines section](https://cfpb.github.io/design-system/foundation/photography#guidelines). @contolini @anselmbradford

PR #1240, opened 2021-06-03
Expand DS documentation tables to full width

contolini:
Editors of the DS website use markdown tables to organize content side-by-side. The HTML tables generated by the markdown are often unreasonably narrow because none of their inner content has a width specified.

I'm not modifying our .o-table organism because I only want to force tables to be full-width on the DS website, and not across cf.gov.

Fixes https://github.com/cfpb/design-system/issues/1239

## Testing

1. The [landing pages page](https://cfpb.github.io/design-system/pages/landing-pages) of the PR preview should have non-squished tables.

### DIFF
--- a/docs/assets/css/variation.less
+++ b/docs/assets/css/variation.less
@@ -36,9 +36,11 @@
         padding-left: 1rem;
         padding-right: 1rem;
     }
+
+    table {
+        .u-w100pct();
+    }
 }
-
-
 
 .m-variation {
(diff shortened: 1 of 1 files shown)


anselmbradford (approved):
Tables are de-squished.

While you're in there, want to get rid of this gap https://github.com/cfpb/design-system/blob/main/docs/assets/css/variation.less#L41-L42 ?

PR #1257, opened 2021-07-09
Maintain text input modifier colors on focus/hover

contolini:
The pacific focus state of our text inputs overrides any modification
colors for error/warning/success. This means that if an input field
is modified while the user is actively focused on it, they won't receive
visual feedback that something has changed.

## Additions

- Hover/focus states to text input modifiers.

## Testing

1. Compare the three production [modified text inputs](https://cfpb.github.io/design-system/components/notifications#warning-notification-field-level) to the [PR preview](https://deploy-preview-1257--cfpb-design-system.netlify.app/design-system/components/notifications#warning-notification-field-level). Also compare them to the [default text inputs](https://cfpb.github.io/design-system/components/text-inputs).

## Notes

-

## Todos

-

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/packages/cfpb-forms/src/atoms/text-input.less
+++ b/packages/cfpb-forms/src/atoms/text-input.less
@@ -45,11 +45,15 @@
         border-color: @input-border__error;
         outline: 1px solid @input-border__error;
         &:hover,
-        &.hover,
+        &.hover {
+            border-color: @input-border__error;
+            outline: 1px solid @input-border__error;
+        }
         &:focus,
         &.focus {
             border-color: @input-border__error;
             box-shadow: 0 0 0 1px @input-border__error;
+            outline: 1px dotted @input-border__error;
         }
     }
 
@@ -57,11 +61,15 @@
         border-color: @input-border__warning;
         outline: 1px solid @input-border__warning;
         &:hover,
-        &.hover,
+        &.hover {
+            border-color: @input-border__warning;
+            outline: 1px solid @input-border__warning;
+        }
         &:focus,
         &.focus {
             border-color: @input-border__warning;
             box-shadow: 0 0 0 1px @input-border__warning;
+            outline: 1px dotted @input-border__warning;
         }
     }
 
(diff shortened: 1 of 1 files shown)


contolini:
TBH I'm not sure I like this change because there's now no visual feedback for hovering... hmmmmmm...

niqjohnson:
@contolini: Yeah, I agree, not having a hover state on fields with errors feels off. What if we brought the pacific hover border back for fields with errors but made sure that the focus state has priority over that, so if you're hovering on a focused field with an error the field will have the error-color border and outline. That's how error-free fields behave, so it'd be consistent, and it would also work for the use case I'm working on where a field needs to show an error state after you type a certain number of characters (so you might be both focused and hovering on the field when it needs to turn into a field with an error).

niqjohnson:
@contolini, what do you think of the commit I just pushed, d039cc5? That brings back regular hover styles to error fields but gives them a focus state that uses the error border color even when you're focused and hovered.

contolini:
@niqjohnson I dig it! Looks good to me.

niqjohnson:
@contolini, awesome. I feel like I probably shouldn't approve the PR now that I worked on it. Up to @anselmbradford?

anselmbradford (approved):
We might want to make a [mixin](https://lesscss.org/features/#mixins-feature) in this less file that takes a color and spits out these generated patterns. Feel free to do that if you think it's a good idea.

contolini:
Thanks @anselmbradford and @niqjohnson! I agree that a mixin makes sense but I'm going to skip it because I'm short on time nowadays and the Less will transpile to the same amount of CSS with or without the mixin so I can sleep guilt-free.

PR #1258, opened 2021-07-12
Fix field-level error message margins

niqjohnson:
Field-level error messages are meant to have a [15px margin between the field with the error and the message](https://cfpb.github.io/design-system/components/notifications#error-notification-field-level). That margin was only present [when the error message was in an `.m-form-field container`](https://github.com/cfpb/design-system/blob/main/packages/cfpb-forms/src/molecules/form-fields.less#L241-L244). Our [text input with button pattern](https://cfpb.github.io/design-system/components/text-inputs#text-input-with-a-button) (where the text field and the button are on the same line) don't have the `.m-form-field` wrapper, so error messages in those kinds of forms had no top margin.

## Additions

- Add top margin to field-level notifications in text input with button forms

## Testing

Whip up a test case that puts a field-level error in a text input with button form, like so:

```html
<div class="o-form__input-w-btn">
    <div class="o-form__input-w-btn_input-container">
        <input class="a-text-input a-text-input__error" type="text" title="Test input">
        <div class="a-form-alert a-form-alert__error" id="form-input-error_message" role="alert">
            {{ svg_icon('error-round') }}
            <span class="a-form-alert_text">
                This is an inline alert with an error state.
            </span>
        </div>
    </div>
    <div class="o-form__input-w-btn_btn-container">
        <button class="a-btn">Search</button>
    </div>
</div>
```
Without this fix, the error message will be flush against the bottom of the input. With the fix, there should a be a 15px margin between the two. See the screenshots below to see how everything should look.

## Screenshots

Before | After
---- | ----
before | after

## Checklist

- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [x] Project documentation has been updated
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Browsers

- [x] Chrome on desktop
- [x] Firefox
- [x] Safari on macOS
- [x] Edge
- [x] Internet Explorer 9, 10, and 11
- [x] Safari on iOS
- [x] Chrome on Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [x] Is useable without CSS
- [x] Is useable without JS
- [x] Flexible from small to large screens
- [x] No linting errors or warnings
- [x] JavaScript tests are passing

### DIFF
--- a/packages/cfpb-forms/src/molecules/form-fields.less
+++ b/packages/cfpb-forms/src/molecules/form-fields.less
@@ -238,6 +238,9 @@
         }
     }
 
+    // TODO: The same top margin is applied to field-level errors for input-
+    // with-button forms in organisms/form.less; we should find a way to merge
+    // these to be less repetitive.
     .a-form-alert,
     .a-error-message {
         margin-top: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
--- a/packages/cfpb-forms/src/organisms/form.less
+++ b/packages/cfpb-forms/src/organisms/form.less
@@ -43,6 +43,9 @@
                 width: 100%;
             }
 
+            // TODO: The same top margin is applied to field-level errors for
+            // regular forms in molecules/form-fields.less; we should find a way
+            // to merge these to be less repetitive.
             .a-form-alert,
             .a-error-message {
                 margin-top: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
--- a/packages/cfpb-forms/src/organisms/form.less
+++ b/packages/cfpb-forms/src/organisms/form.less
@@ -42,6 +42,11 @@
                 box-sizing: border-box;
                 width: 100%;
             }
+
+            .a-form-alert,
+            .a-error-message {
+                margin-top: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
+            }
         }
 
         &_btn-container {


anselmbradford:
Instead of duplicating this block twice ([here](https://github.com/cfpb/design-system/blob/main/packages/cfpb-forms/src/molecules/form-fields.less#L241-L244) and in this PR), could we put it in a place that would hit both situations? For instance, would all occurrences be inside an `o-form`? 

```
.o-form {
    .a-form-alert,
    .a-error-message {
        margin-top: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
    }
}
```

… and get rid of the other two spots?

niqjohnson:
@anselmbradford: Yeah, I was looking for a way to be less repetitive with the margins but couldn't find anything super obvious. I ran into a couple of problems.

The [sample code on the Design System](https://cfpb.github.io/design-system/components/notifications#error-notification-field-level) has only an `m-form-field` as a container, so there's a possibility that notifications are out there in the wild using that pattern, too (i.e. not necessarily wrapped in any of our `o-form` classes). I did a quick search of the cf.gov codebase and saw only a [couple of uses of `a-form-alert`](https://github.com/cfpb/consumerfinance.gov/search?q=a-form-alert), neither of which is using the same pattern that's in the Design System docs, so maybe this isn't a big concern.

The bigger problem is that we seem to be pretty inconsistent with what form containers we use. I took a quick look at a few of our more common forms, and there doesn't seem to be a common container pattern: the feedback form uses `o-form`, the filter form doesn't have a top-level container but does use `o-form_group`s, and the Ask search form uses `o-form__input-w-btn`. I don't think any of those forms have field-level notifications, and they probably all *should* use an `o-form` container, but right now there doesn't seem to be a consistent container we could use in the selector. The best I could come up with is

```css
[class^=o-form] {
    .a-form-alert,
    .a-error-message {
        margin-top: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
    }
}
```

It's less repetitive, but is it more elegant?

anselmbradford (approved):
Hmm thanks for looking into it. Maybe add a TODO comment about removing the redundancy.

niqjohnson:
Sounds good @anselmbradford, thanks. I added the TODO note in both places.

PR #1259, opened 2021-07-21
Adjust mobile nav margin and link underlining

brandon-rusk:
Draft PR for Mobile Nav tweaks -- current known issues:
-ALL links are now solid borders. I believe it would only make sense to make mobile nav links solid borders to match cf.gov. For consistency, all links on the actual pages should be dotted like cf.gov (?)
-There are a top AND bottom border on the mobile nav links. Should change this to only be a bottom border

### DIFF
--- a/docs/assets/css/sidebar.less
+++ b/docs/assets/css/sidebar.less
@@ -46,6 +46,11 @@
 // Mobile size.
 .respond-to-max( @bp-xs-max, {
     
+    a {
+        border-style: solid !important;
+        border-top-width: 0px !important;
+    }
+
     .ds-nav-2 .ds-nav-3 {
         margin-top: 15px;
         margin-left: 10px;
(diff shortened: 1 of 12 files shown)


docs/assets/css/sidebar.less
```diff
@@ -50,12 +45,21 @@
...
-    .ds-nav .h4 {
-        margin-bottom: 0;
+    
+    .ds-nav a {
+        border-style: solid;
+        border-top-width: 0;
     }
 
-    .ds-nav-2 {
+    .ds-nav-2 .ds-nav-3 {
```
anselmbradford, line 54: Are you trying to target both `ds-nav-2` and `ds-nav-3`? If it's both of them, you'll need a comma 

```
.ds-nav-2,
.ds-nav-3 {
```

brandon-rusk: Nice catch, thank you. Fixed.

anselmbradford (approved):
LGTM 👍 You could squash and merge (drop-down arrow on the merge button) to squash the commits together.

PR #1260, opened 2021-07-23
Fix images displaying in a squished manner

brandon-rusk:
Images no longer are squished.

Before:
image

After:
image

Fixes https://github.com/cfpb/design-system/issues/1239


docs/assets/css/variation.less
```diff
@@ -86,3 +82,8 @@
 .m-list__horizontal-spaced > .m-list_item {
     margin-bottom: unit( 15px / @base-font-size-px, em );
 }
+
+// Gives 100% width to all tables
+table {
+    .u-w100pct();
+}
```
anselmbradford, line 89: TRIVIAL: Let's add the ending newline to remove the warning.

contolini:
When I originally added that `.u-w100pct();` line I failed to realized that it messes with the [live table component examples](https://cfpb.github.io/design-system/components/tables#standard-table-1).

@brandon-rusk How would you feel about adding a second CSS rule to this PR with something like:

```
.a-live_code table {
    width: inherit;
}
```

The goal would be to expand every table on the website *except* ones that are in code examples.

brandon-rusk:
> When I originally added that `.u-w100pct();` line I failed to realized that it messes with the [live table component examples](https://cfpb.github.io/design-system/components/tables#standard-table-1).
> 
> @brandon-rusk How would you feel about adding a second CSS rule to this PR with something like:
> 
> ```
> .a-live_code table {
>     width: inherit;
> }
> ```
> 
> The goal would be to expand every table on the website _except_ ones that are in code examples.

Hi Chris, I want to confirm what the functionality here is supposed to be. We want the live tables to be full width, correct?

With just the `table {
    .u-w100pct();
}`:
image


With `.a-live_code table {
    width: inherit;
}` added:
image

contolini:
We want the live code tables to *not* be full width because we want them to be representative of only the styles in the [tables component](https://github.com/cfpb/design-system/blob/main/packages/cfpb-tables/src/cfpb-tables.less) and not any of the [docs website styles](https://github.com/cfpb/design-system/tree/main/docs/assets/css).

It's confusing because they look like hot trash when they're not full width but its an accurate representation of the table component's source code.

brandon-rusk:
> We want the live code tables to _not_ be full width because we want them to be representative of only the styles in the [tables component](https://github.com/cfpb/design-system/blob/main/packages/cfpb-tables/src/cfpb-tables.less) and not any of the [docs website styles](https://github.com/cfpb/design-system/tree/main/docs/assets/css).
> 
> It's confusing because (as demonstrated by your screenshots) they look like hot squished trash when they're not full width but its an accurate representation of the table component's source code.

Understood, thank you for the clarification! I will update this immediately.

anselmbradford (approved):

PR #1263, opened 2021-08-02
Remove govuk tabs dependency and roll our own

brandon-rusk:
## Changes

- Remove govuk tabs dependency and create our own Tabs component. 

## Testing

1. PR checks should pass.
2. Check PR preview and see that the tabs under the "SHOW DETAILS" toggle work.
3. Edit a page and see that the tabs work in the admin preview.

## Todo

- The tabs CSS should be tweaked to fix https://github.com/cfpb/design-system/issues/1276


docs/_includes/variation-content.html
```diff
@@ -131,32 +131,32 @@ <h3>{{ variation_name }}</h3>
                 <button href="#{{ variation_slug }}-{{ forloop.index }}" class="a-btn a-btn__link u-hidden" data-toggle-details="hide">Hide details</button>
             </div>
 
-            <div class="govuk-tabs u-hidden" data-module="tabs" id="{{ variation_slug }}-{{ forloop.index }}">
-                <ul class="govuk-tabs__list">
+            <div class="m-tabs u-hidden" data-module="tabs" id="{{ variation_slug }}-{{ forloop.index }}">
+                <ul class="m-tabs__list">
```
anselmbradford, line 135: In our flavor of BEM, the syntax is `block_element__modifier`, so this is saying the `__list` modifier on the `m-tabs` block. However, modifiers should appear with their block, so `m-tabs m-tabs__list`. 

A theoretical example is think of a red square block where we modify it to change the background color to blue. `m-square` is the default red square and `m-square m-square__blue` is the modified blue version.

So instead of the `m-tabs__list` modifier, this would be an `m-tabs_list` element. This is an unfortunate mapping. Govuk actually also uses BEM, but for whatever historical reason our version diverged from the official BEM syntax :( Govuk's element designator is `__` and their modifier is `--`, while our element is `_` and modifier is `__`. 

So bottom line is this is an element in both govuk land and ours, but our syntax to specify it as such differs.

docs/_includes/variation-content.html
```diff
@@ -177,7 +177,7 @@ <h4 class="variation-code-snippet-title">HTML code snippet</h4>
 
                 <!-- Jinja snippet -->
                 {% if jinja_snippet != "" %}
-                <section class="govuk-tabs__panel govuk-tabs__panel--hidden" id="{{ variation_slug }}-jinja-code-snippet-{{ forloop.index }}">
+                <section class="m-tabs__panel m-tabs__panel--hidden" id="{{ variation_slug }}-jinja-code-snippet-{{ forloop.index }}">
```
anselmbradford, line 180: Here they're putting a modifier on the `panel` element. We don't put modifiers on elements much or at all, so it's discouraged, but if it makes it easier to migrate this over it's fine. However, I think we'll want to use `u-hidden` (utility class) to handle the hide/show as opposed to a modifier.

docs/assets/css/tabs.less
```diff
@@ -1,166 +1,103 @@
...
-    .govuk-tabs {
-        margin: 0 -1rem;
-    }
-}
-
-.govuk-tabs__list {
-    border-top: 1px solid @gray;
+.m-tabs {
+    width: 20em;
+    font-family: "lucida grande", sans-serif;
```
anselmbradford, line 3: Probably don't want to use this font-family. Where's this coming from? Is that the govuk default? I'd just leave `font-family` out altogether unless it looks disasterous.

docs/assets/css/tabs.less
```diff
@@ -1,166 +1,103 @@
...
+  }
+  
+  [role="m-tabs__list"] {
```
anselmbradford, line 6: `role` will be an [aria role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques) from a predefined set, so we can't arbitrarily set it to a class name.

In the [W3 example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html) they have `role="tab"` and doing the matching on that because they don't have class names defined. You do have class names so can do the matching on that, so `.m-tabs_list {` (note I've converted it into an element syntax)

docs/assets/css/tabs.less
```diff
@@ -1,166 +1,103 @@
...
-.variation-jinja-code-snippet-title,
-.variation-implementation-title,
-.variation-specs-title {
-    .respond-to-min( 641px, {
-        .u-visually-hidden();
-    } );
-    background-color: @white;
-    border-bottom: 1px solid @gray;
+    padding: 0.3em 0.5em 0.4em;
+    border: 1px solid hsl(219, 1%, 72%);
```
anselmbradford, line 15: Are the `hsl` copied values from the web inspector? We have a set of [color variables](https://github.com/cfpb/design-system/blob/main/docs/assets/css/color-swatches.less#L77) (swatch examples [here](https://cfpb.github.io/design-system/foundation/color)) that we'll want to use. Looking at the prior code looks like this would be `@gray`

Same thing with the background color. Refer to the swatches and https://github.com/cfpb/design-system/blob/main/docs/assets/css/color-swatches.less#L77 for the color variable name that's appropriate.

docs/assets/css/tabs.less
```diff
@@ -1,166 +1,103 @@
...
+    right: -1px;
+    left: -1px;
+    border-radius: 0.2em 0.2em 0 0;
+    border-top: 3px solid hsl(20, 96%, 48%);
+    content: '';
+  }
+  
+  [role="m-tabs__tab"][aria-selected="true"] {
+    border-radius: 0;
+    background: hsl(220, 43%, 99%);
```
anselmbradford, line 38: Use color variable.

docs/assets/css/tabs.less
```diff
@@ -1,166 +1,103 @@
...
+    outline: 0;
+  }
+  
+  [role="m-tabs__tab"][aria-selected="true"]:not(:focus):not(:hover)::before {
+    border-top: 5px solid hsl(218, 96%, 48%);
```
anselmbradford, line 43: Use color variable.

docs/assets/css/tabs.less
```diff
@@ -1,166 +1,103 @@
...
+  }
+  
+  [role="m-tabs__tab"][aria-selected="true"]::after {
+    position: absolute;
+    z-index: 3;
+    bottom: -1px;
+    right: 0;
+    left: 0;
+    height: 0.3em;
+    background: hsl(220, 43%, 99%);
```
anselmbradford, line 53: Use color variable.

docs/assets/css/tabs.less
```diff
@@ -1,166 +1,103 @@
...
+  [role="m-tabs__tab"]:focus,
+  [role="m-tabs__tab"]:active {
+    outline: 0;
+    border-radius: 0;
+    color: inherit;
+  }
+  
+  [role="m-tabs__tab"]:hover::before,
+  [role="m-tabs__tab"]:focus::before {
+    border-color: hsl(20, 96%, 48%);
```
anselmbradford, line 68: Use color variable.

docs/assets/css/tabs.less
```diff
@@ -1,166 +1,103 @@
...
+  }
+  
+  [role="m-tabs__panel"] {
+    position: relative;
+    z-index: 2;
+    padding: 0.5em 0.5em 0.7em;
+    border: 1px solid hsl(219, 1%, 72%);
```
anselmbradford, line 75: Use color variable.

docs/assets/css/tabs.less
```diff
@@ -1,166 +1,103 @@
...
+    border-radius: 0 0.2em 0.2em 0.2em;
+    box-shadow: 0 0 0.2em hsl(219, 1%, 72%);
```
anselmbradford, line 77: Use color variable.

docs/assets/css/tabs.less
```diff
@@ -1,166 +1,103 @@
...
+    background: hsl(220, 43%, 99%);
```
anselmbradford, line 78: Use color variable.

docs/assets/css/tabs.less
```diff
@@ -1,166 +1,103 @@
...
+  }
+  
+  [role="m-tabs__panel"]:focus {
+    border-color: hsl(20, 96%, 48%);
```
anselmbradford, line 82: Use color variable.

docs/assets/css/tabs.less
```diff
@@ -1,166 +1,103 @@
...
+    box-shadow: 0 0 0.2em hsl(20, 96%, 48%);
```
anselmbradford, line 83: Use color variable.

docs/assets/css/tabs.less
```diff
@@ -1,166 +1,103 @@
...
+    outline: 0;
+  }
+  
+  [role="m-tabs__panel"]:focus::after {
+    position: absolute;
+    bottom: 0;
+    right: -1px;
+    left: -1px;
+    border-bottom: 3px solid hsl(20, 96%, 48%);
```
anselmbradford, line 92: Use color variable.

docs/assets/js/main.js
```diff
@@ -9,7 +9,7 @@ import AlphaTransition from '@cfpb/cfpb-atomic-component/src/utilities/transitio
 import MoveTransition from '@cfpb/cfpb-atomic-component/src/utilities/transition/MoveTransition.js';
 import MaxHeightTransition from '@cfpb/cfpb-atomic-component/src/utilities/transition/MaxHeightTransition.js';
 import Table from '@cfpb/cfpb-tables/src/Table';
-import { Tabs } from 'govuk-frontend';
+import Tabs from './tabs'
```
anselmbradford, line 12: We're obviously not consistent with this, but sometimes I think it's a good idea to add the `.js` because then you can search for the file name and see where it's imported.

docs/assets/js/tabs.js
```diff
@@ -1,52 +1,259 @@
...
-  const tabs = container.querySelectorAll( '[data-module="tabs"]' );
-
-  if ( tabs && tabs.length > 0 ) {
-    for ( let i = 0; i < tabs.length; i++ ) {
-      const tab = tabs[i];
-      new Tabs( tab ).init();
+const BASE_CLASS = 'm-tabs';
+
+function Tabs() {
+  const tablist = document.querySelectorAll('[role="m-tabs__list"]')[0];
```
anselmbradford, line 4: Search for the tab class inside the base class, so have something like:

```
const tabsContainerDom = document.querySelector( `.${ BASE_CLASS }` );
const tabsListDom = tabsContainerDom.querySelectorAll( `.${ BASE_CLASS }_list` );
```

Also, both these lookups I'd put in the init method, so:

```
let tabsContainerDom;
let tabsListDom;

function init() {
  const tabsContainerDom = document.querySelector( `.${ BASE_CLASS }` );
  const tabsListDom = tabsContainerDom.querySelectorAll( `.${ BASE_CLASS }_list` );

  return this;
}
```

This is just. so when you create the instance, it creates the internal. variables, but they don't get assigned a value till `init` is called.

docs/assets/js/tabs.js
```diff
@@ -1,52 +1,259 @@
...
+  let tabs;
+  let panels;
+  // const delay = determineDelay();
+
+  generateArrays();
```
anselmbradford, line 9: Should go inside `init()`

docs/assets/js/tabs.js
```diff
@@ -1,52 +1,259 @@
...
+  // Add or substract depending on key pressed
+  const direction = {
+    37: -1,
+    38: -1,
+    39: 1,
+    40: 1
+  };
+
+  // Bind listeners
+  for (let i = 0; i < tabs.length; ++i) {
```
anselmbradford, line 36: Should go inside init or it's own `_bindListeners()` private method. Remember the `_` prefix just means a method that is not exposed externally by the class at the end of the file with this.method = method.

docs/assets/js/tabs.js
```diff
@@ -1,52 +1,259 @@
...
+
+  // Only activate tab on focus if it still has focus after the delay
+  function checkTabFocus (target) {
+    focused = document.activeElement;
+
+    if (target === focused) {
+      activateTab(target, false);
+    };
+  };
 }
```
anselmbradford, line 257: We'll need to expose the public API for the Tabs class and return the instance, like https://github.com/cfpb/design-system/blob/main/packages/cfpb-forms/src/organisms/Multiselect.js#L559-L570

contolini (approved):
Very excited to remove this cruft--nice work!

PR #466, opened 2020-04-29
Migrate pages to new variation group structure

contolini:
This moves every component page to the new "group" format that we've been experimenting with over at https://github.com/cfpb-sandbox/design-system. It involves a heckuva lot of changes because the frontmatter of every markdown file has been slightly altered to accommodate the new format.

## Changes

- Replaces the `variations` field of every collection with a `variation_groups` field.
- Updates every markdown file to use the new `variation_groups` field.
- Replaces "Show/hide code & specs" button that was at the top of every page with a "Show/hide source" link at the bottom of each variation.

## Testing

1. Click around the [PR preview](https://deploy-preview-466--cfpb-design-system.netlify.app/design-system/components/) and look for obvious signs of wonkiness.
1. The browser tests have been updated and should pass: `yarn test:browser`

## Screenshots

| notifications page before | notifications page after |
|--------|-------|
| Screen Shot 2020-04-29 at 2 35 39 PM | Screen Shot 2020-04-29 at 2 35 51 PM |

## Notes

-

## Todos

-

## Checklist

- [ ] PR has an informative and human-readable title
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code can be automatically merged (no conflicts)
- [ ] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [ ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Browsers

- [x] Chrome on desktop
- [x] Firefox
- [x] Safari on macOS
- [x] Edge
- [x] Internet Explorer 9, 10, and 11
- [x] Safari on iOS
- [x] Chrome on Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [x] Is useable without CSS
- [x] Is useable without JS
- [x] Flexible from small to large screens
- [ ] No linting errors or warnings
- [ ] JavaScript tests are passing

PR #1266, opened 2021-08-16
Anchor linking: Fix live code example

anselmbradford:
https://github.com/cfpb/design-system/pull/443/ exempted live code examples from anchor linking, but targets `live-code-example` while the live code blocks use `a-live_code` as changed in https://github.com/cfpb/design-system/pull/466/. Erroneous links appear within the code examples as seen on https://cfpb.github.io/design-system/patterns/cards

Screen Shot 2021-08-16 at 9 33 20 AM

Screen Shot 2021-08-16 at 9 38 43 AM

## Changes

- Change anchor linking to fix class name.

## Testing

1. Visit the cards pattern and see that there is no anchor link in the links within the card patterns.


contolini (approved):
The world needs an `h*` selector.

Issue #1270, opened 2021-08-16
Unused variation CSS


anselmbradford:
**Describe the bug**

We have some variation classes that appear unused in https://github.com/cfpb/design-system/blob/main/docs/assets/css/variation.less, specifically:

`t-variation`
`variation-code-specs`

anselmbradford:
https://github.com/cfpb/design-system/pull/1285 is merged

PR #1285, opened 2021-09-02
Variation Cleanup

### LINKED ISSUE #1269: o-variation_group appears to be misnamed
**Describe the bug**

We have `o-variation_group` in the CSS https://github.com/cfpb/design-system/blob/main/docs/assets/css/variation.less#L20, but in the template we don't have `o-variation`, only `m-variation`, so `o-variation_group` is the Block in BEM and should be `o-variation-group`.

brandon-rusk:
Removed unused code and fixed some naming conventions to resolve #1269 and #1270

### DIFF
--- a/docs/_includes/variation-content.html
+++ b/docs/_includes/variation-content.html
@@ -59,10 +59,10 @@
 
     {% if page.variation_groups and page.variation_groups.size > 0 %}
     {% for variation_group in page.variation_groups %}
-    <section class="o-variation_group" id="{{ variation_group.variation_group_name | slugify }}">
+    <section class="o-variation-group" id="{{ variation_group.variation_group_name | slugify }}">
 
         {% if variation_group.variation_group_name and variation_group.variation_group_name != '' %}
-        <h2 class="o-variation_group_title">
+        <h2 class="o-variation-group-title">
             {{ variation_group.variation_group_name }}
         </h2>
         {% else %}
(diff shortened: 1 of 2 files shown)


docs/_includes/variation-content.html
```diff
@@ -59,10 +59,10 @@ <h1 class="component-header u-mb30">
 
     {% if page.variation_groups and page.variation_groups.size > 0 %}
     {% for variation_group in page.variation_groups %}
-    <section class="o-variation_group" id="{{ variation_group.variation_group_name | slugify }}">
+    <section class="o-variation-group" id="{{ variation_group.variation_group_name | slugify }}">
 
         {% if variation_group.variation_group_name and variation_group.variation_group_name != '' %}
-        <h2 class="o-variation_group_title">
+        <h2 class="o-variation-group-title">
```
anselmbradford, line 65: So in BEM this would actually be `o-variation-group_title` where `o-variation-group` is the Block (B) and `_title` is the Element (E) of that block.

docs/_includes/variation-content.html
```diff
@@ -74,7 +74,7 @@ <h2 class="o-variation_group_title">
         {% endif %}
 
         {% if variation_group.variation_group_description %}
-        <div class="o-variation_group_description">
+        <div class="o-variation-group-description">
```
anselmbradford, line 77: Same as above, `o-variation-group` is the block, and description is the element`, so it would be `o-variation-group_description`.

docs/_includes/variation-content.html
brandon-rusk: Ahh, I was wondering about this...
Updated, thanks!

docs/_includes/variation-content.html
brandon-rusk: Updated!

anselmbradford (approved):

PR #1296, opened 2021-09-20
Fix multiselect infinite selections bug

anselmbradford:
It was possible to select infinite options in the multiselect if the multiselect was loaded with the max selection count. This PR fixes that loophole.

## Changes

- Changes multiselect model to stop checking options after the maximum selection number has been reached.
- Changes multiselect to initialize the model before populating the markup and to check the model for the selections.
- Updates the multiselect example in the DS docs to load with the maximum selection to begin with.
 
## Testing

1. Check the multiselect in the PR preview and add and remove items to see that it works.
2. If you really want to thoroughly vet this, pull the branch and run a local server and change the number of items in dropdowns.md that have the `selected` attribute so-as to change the number of items the multiselect has on page load.

### DIFF
--- a/docs/pages/dropdowns.md
+++ b/docs/pages/dropdowns.md
@@ -126,12 +126,12 @@ variation_groups:
               </label>
               <select class="o-multiselect" id="test_select__multiple" multiple>
                   <option value="option1" selected>Option 1</option>
-                  <option value="option2">Option 2</option>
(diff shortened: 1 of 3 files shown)


anselmbradford:
See https://[GHE]/CFGOV/platform/issues/4082

brandon-rusk (approved):
LGTM!

PR #1301, opened 2021-09-23
Add taglines component

anselmbradford:
## Additions

- Adds a tagline page.
- Adds a tagline example to the inkwell.

## Changes

- Refactors the tagline atom to:
 - make the flag a utility with an inlined image.
 - make an element for the US government text so it doesn't wrap.
 - use CSS grid for layout.

## Testing

1. Check out the tagline and wells page in the PR preview.

## Screenshots
tagline

Screen Shot 2021-09-23 at 1 55 55 PM


brandon-rusk (approved):
LGTM!

packages/cfpb-typography/src/atoms/tagline.less
```diff
@@ -0,0 +1,52 @@
...
+    display: grid;
+    grid-template-columns: 22px 1fr;
+    grid-column-gap: 10px;
+
+    // This'll target the USA flag.
+    & > *:first-child {
+        margin-top: 1px;
+    }
+
+    &_usa {
```
contolini, line 11: Should this be a reusable utility class similar to `.u-clearfix`(e.g. `.u-nowrap` )?

docs/pages/taglines.md
```diff
@@ -0,0 +1,59 @@
...
+          The flag itself is a stand-alone element of
+          `<span class="u-usa-flag"></span>`
+          that uses a utility class that embeds a double-resolution flag png
+          via a data URI.
+      - variation_name: "Large tagline"
+        variation_code_snippet: >-
+          <div class="a-tagline a-tagline__large">
+              <span class="u-usa-flag"></span>
+              <div>
+                An official website of the
```
contolini, line 30: Totally trivial but it's annoying how netlify CMS seems to mix two- and four-space indentations.

packages/cfpb-typography/src/atoms/tagline.less
```diff
@@ -0,0 +1,52 @@
+.a-tagline {
+    display: grid;
+    grid-template-columns: 22px 1fr;
+    grid-column-gap: 10px;
+
+    // This'll target the USA flag.
+    & > *:first-child {
```
contolini, line 7: Is this wildcard preferred over specifically targeting the flag element? Is this because there *could* be a different non-flag element that comes first?

contolini (approved):
Looks good 👍 Just a couple questions.

packages/cfpb-typography/src/atoms/tagline.less
anselmbradford: I was just making it generic that way, but I guess that's highly unlikely and it would probably render ever so quicker without a wildcard.

Fixed in https://github.com/cfpb/design-system/pull/1301/commits/4f615970bd6ccf87bd1bf3f5983b135b60cdc131

packages/cfpb-typography/src/atoms/tagline.less
anselmbradford: MMmmmmmm I like it. Fixed in https://github.com/cfpb/design-system/pull/1301/commits/4f615970bd6ccf87bd1bf3f5983b135b60cdc131

anselmbradford: Now the tagline is just a container for arbitrary text content, which may contain markup and a `no-wrap` class. I like it better.

Issue #1300, opened 2021-09-23
Highlight card links have insufficient contrast ratio


chosak:
**Describe the bug**

Cards page fails Lighthouse accessibility check due to insufficient contrast ratio.

Lighthouse reports this error: "Background and foreground colors do not have a sufficient contrast ratio." See most recent nightly Lighthouse run [here](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1632356824126-31055.report.html), as run by GitHub Actions [here](https://github.com/cfpb/design-system/runs/3681684398?check_suite_focus=true).

Foreground color is #0072ce. Background color is #e2efd8. Contrast ratio is 4.08. WCAG 2.0 AA requires a contrast ratio of 4.5.

**To Reproduce**

1. Go to https://cfpb.github.io/design-system/patterns/cards in Chrome.
2. Run a Lighthouse accessibility test.
3. See error.

**Expected behavior**

WCAG 2.0 AA requires a contrast ratio of 4.5.

**Screenshots**

Current appearance of highlight card pattern:

![image](https://user-images.githubusercontent.com/654645/134533866-289dd656-3f27-480d-b6f4-454266c42c71.png)

Error as reported in Lighthouse:

![image](https://user-images.githubusercontent.com/654645/134532624-98100536-893b-41d8-9a8e-44df3ddd2706.png)

PR #1302, opened 2021-09-23
Cards: Make highlight card colors accessible

anselmbradford:
Fixes https://github.com/cfpb/design-system/issues/1300

## Changes

- Adds a utility for setting all the states of the cards and darkens the highlight card link text to be an accessible contrast.

## Testing

1. Visit the cards page in the PR preview and check for contrast accessibility errors.

### DIFF
--- a/packages/cfpb-layout/src/molecules/card.less
+++ b/packages/cfpb-layout/src/molecules/card.less
@@ -20,6 +20,10 @@
         color: @v;
     }
 
+    // Border changes on the regular cards happen on the top-level `article`
+    // element, so for consistency we trigger the hover styles on the parent
+    // instead of on the link. This differs from the visited, focus,
+    // and active states, which add styles onto the link.
     &:hover .m-card_footer > span {
         border-style: solid;
         border-color: @h;
--- a/packages/cfpb-layout/src/molecules/card.less
+++ b/packages/cfpb-layout/src/molecules/card.less
@@ -1,25 +1,38 @@
-.u-card-link-text() {
+// @c = default state.
+// @v = `:visited` state.
+// @h = `:hover` state.
+// @f = `:focus` state.
+// @a = `:active` state.
+.u-link-card__colors( @c, @v, @h, @f, @a ) {
     .m-card_footer > span {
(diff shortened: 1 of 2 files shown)


packages/cfpb-layout/src/organisms/wells.less
```diff
@@ -19,7 +19,7 @@
         color: @white;
 
         a {
-            // default, `:visited`, `:hover`, `:focus`, and `:active` states in the first five arguments)
+            // default, `:visited`, `:hover`, `:focus`, and `:active` states are the first five arguments.
```
chosak, line 22:
```suggestion
            // Arguments: default, `:visited`, `:hover`, `:focus`, and `:active` states.
```

Just to make it consistent with the other lines.

packages/cfpb-layout/src/molecules/card.less
```diff
@@ -1,25 +1,38 @@
-.u-card-link-text() {
+// @c = default state.
+// @v = `:visited` state.
+// @h = `:hover` state.
+// @f = `:focus` state.
+// @a = `:active` state.
+.u-link-card__colors( @c, @v, @h, @f, @a ) {
```
chosak, line 6: Would it make more sense to use `@d` for default?

chosak (commented):
image

Nice work on the accessibility fix, but one question: is the focus/active color working properly? When I highlight the card the color doesn't seem to change, and the link itself isn't keyboard-focusable.

packages/cfpb-layout/src/molecules/card.less
anselmbradford: I was following the format of `u-link__colors-base`, which uses `@c` https://github.com/cfpb/design-system/blob/main/packages/cfpb-core/src/utilities.less#L187

Probably stands for "current"

anselmbradford:
Note I made the hover, focus, and active states the same, but we may want to differentiate those at some point (focus gets the focus rectangle, so it's really hover and active that maybe should be differentiated).

chosak (approved):

PR #1305, opened 2021-10-04
Taglines: Move stacking to only affect extra-large size

anselmbradford:
Preference is to have the tagline stack under 600px only when the extra-large size is used.

## Changes

- Taglines: Move stacking to only affect xtra-large size

## Testing

1. Visit taglines PR preview page and resize to mobile and see that only the extra-large size stacks with the flag on its own line.

## Screenshot

Screen Shot 2021-10-04 at 12 49 52 PM

### DIFF
--- a/docs/pages/taglines.md
+++ b/docs/pages/taglines.md
@@ -43,7 +43,7 @@ variation_groups:
               We're a government agency whose mission is to protect consumers
               from financial harm.
           </div>
-        variation_description: ""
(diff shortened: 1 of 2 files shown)


contolini (approved):
Thxxx the stacking was bothering me

chosak (approved):

brandon-rusk (approved):
LGTM

PR #1330, opened 2021-11-03
Tagline: Adjust font-size, icon-text spacing, and max-width in inkwell

anselmbradford:
## Changes

- Give the tagline a max-width inside inkwells.
- Increase the tagline gap between icon and text at the extra large size.
- Increase the font-size of the tagline at the extra large size.

## Testing

1. Compare PR preview tagline to https://cfpb.github.io/design-system/patterns/taglines#extra-large-tagline and https://cfpb.github.io/design-system/patterns/wells#inkwell

## Screenshots

Before:
Screen Shot 2021-11-03 at 4 06 52 PM
Screen Shot 2021-11-03 at 4 07 05 PM

After
Screen Shot 2021-11-03 at 4 07 18 PM
Screen Shot 2021-11-03 at 4 07 30 PM
:

### DIFF
--- a/packages/cfpb-layout/src/organisms/wells.less
+++ b/packages/cfpb-layout/src/organisms/wells.less
@@ -25,6 +25,7 @@
         // If an inkwell contains a tagline, properly lay it out.
         .a-tagline {
             margin-bottom: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
+            max-width: 41.875rem;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


contolini (approved):
We may want to consider moving the `41.875rem` magic number to a variable in `cfpb-layout.less` (something like `@content-max-width: 41.875rem`).

anselmbradford:
> We may want to consider moving the 41.875rem magic number to a variable in cfpb-layout.less (something like @content-max-width: 41.875rem).

Hmmmmmm not a bad idea. I'll PR it separately.

PR #1342, opened 2021-12-15
Remove empty link from typography page

chosak:
It looks like an empty link was inadvertently added to the typography page in #1341.


caheberer:
Thanks for catching, @chosak

PR #1347, opened 2022-03-17
Pull old ie support

wpears:
We ship css to all our users that is only targeting browsers that are sub-0.5% usage. This is wasteful and blocking for people on slow connections and mobile phones with slower processing speeds (which are a much greater proportion of our users that ie8, ie9, and ie10).

Thus, this removes those classes. There will be another PR coming in https://github.com/cfpb/consumerfinance.gov to match this behavior and remove modernizr (which was only really used by the `.no-csspointerevents` class here).

### DIFF
--- a/packages/cfpb-buttons/src/atoms/buttons-with-icons.less
+++ b/packages/cfpb-buttons/src/atoms/buttons-with-icons.less
@@ -4,13 +4,6 @@
 // TODO: Replace magic numbers with calculations based off of the
 // button padding size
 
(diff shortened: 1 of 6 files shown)


anselmbradford:
> 😎 At an upcoming FEWD bi-weekly we should revisit our browser support best practices.

Wish we had [this file](https://github.com/cfpb/cfpb-analytics/blob/main/packages/browserslist-config/browserslist-stats.json) updated from GA data on an interval. What would need to happen to make that happen?

PR #1362, opened 2022-04-21
Fix percentage parentheses for Less 4.x compatibility

anselmbradford:
Tiny addendum to https://github.com/cfpb/design-system/pull/1358

## Changes

- Add needed missing parentheses to `percentage(…)` mixin.

## Testing

1. `yarn build` should pass.


contolini (approved):
🙇

PR #1374, opened 2022-05-18
Centralize text input with a button pattern in the design system

wpears:
Currently our "text input with a button" pattern looks as follows:
Screen Shot 2022-05-18 at 12 55 09 PM

However, everywhere this is used in cf.gov, we specifically add styles that make the button flush with the input, like this:
Screen Shot 2022-05-18 at 12 51 52 PM

This brings that behavior into the design-system. After merge and release, a follow-on PR to cf.gov will remove all that soon-to-be unneeded customization.

### DIFF
--- a/packages/cfpb-forms/src/organisms/form.less
+++ b/packages/cfpb-forms/src/organisms/form.less
@@ -32,12 +32,6 @@
         border-right-width: 0;
       } );
 
-      .respond-to-min( 960px, {
-        .grid_column( 10 );
-
-        border-right-width: 0;
-      } );
-
       .a-text-input {
         box-sizing: border-box;
         width: 100%;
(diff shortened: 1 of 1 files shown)


packages/cfpb-forms/src/organisms/form.less
```diff
@@ -55,18 +49,18 @@
...
-        .grid_column( 2 );
-      } );
-
       .a-btn {
-        box-sizing: border-box;
         width: 100%;
       }
+
+      .respond-to-min( 480px, {
+        .grid_column( 3 );
```
anselmbradford, line 57: We should migrate away from this, but ok for now

anselmbradford, line 57: We should migrate away from this, but ok for now

anselmbradford (approved):

PR #1388, opened 2022-08-11
Add variation group name to Banners page

chosak:
This fixes an accessibility error that Lighthouse discovered.
Without this change, the heading order goes h1, h3.
This adds the missing h2.

### DIFF
--- a/docs/pages/banners.md
+++ b/docs/pages/banners.md
@@ -70,6 +70,6 @@ variation_groups:
               </div>
           </div>
         variation_name: Dark banner
-    variation_group_name: ""
+    variation_group_name: Types
 last_updated: 2020-01-28T15:55:47.394Z
 ---


chosak:
If you run Lighthouse against the preview link above, it score 100%, unlike the current version which scores 98% and thus fails the nightly GitHub Actions check ([here](https://github.com/cfpb/design-system/runs/7778040834?check_suite_focus=true)).

Granted, this layout doesn't look ideal with _only_ a "Types" heading but I'm not sure of an easier way to easily fix this:

image

contolini (approved):
🎉

sonnakim:
@contolini sorry I know this isn't the right place to ask, but is anyone from your team planning to add use cases to this page, so people know when to use the dark banner ?

contolini:
@sonnakim this is a smart idea. One that I like quite a bit.

PR #1396, opened 2022-09-30
Add a subsection eyebrow

wpears:
This adds a subsection eyebrow as shown below to support desired design of the equity-centered design section
Screen Shot 2022-09-30 at 9 48 38 AM

There was an old, unused `secondary_section` variable (looks like this used to be for handling sidebar rendering? This is now handled elsewhere). I've replaced this with an `eyebrow` variable which refers to the current subsection or third-level section a page is nested within.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,31 +1,3 @@
-Must provide GITHUB_AUTH
-
-
-Must provide GITHUB_AUTH
-
-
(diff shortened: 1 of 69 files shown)


jenn-franklin:
Amazing!! Looks perfect. Thank you!

jenn-franklin (approved):

PR #1404, opened 2022-10-06
Update outdated dependencies | remove legacy babel plugins

anselmbradford:
## Changes

- Update outdated dependencies 
- Remove legacy babel plugins
- Remove unreferenced `escape-html` plugin

## Testing

1. PR checks should pass.


wpears (changes requested):
Can we pin the updated deps to major.minor.patch?

wpears (approved):

Issue #1398, opened 2022-10-03
Expandables move toggle to bottom at narrow screen widths


anselmbradford:
1. Go to https://cfpb.github.io/design-system/components/expandables
2. Resize to a narrowish screen width
3. See that the toggle moves downward from it's normal position (also note if the expandable is open already that the content overflows.

<img width="446" alt="Screen Shot 2022-10-03 at 9 47 47 AM" src="https://user-images.githubusercontent.com/704760/193593858-b2b68c0d-50cf-4bf6-9ba7-3a2c6357f3db.png">

PR #1412, opened 2022-10-11
Convert expandable header to flexbox

anselmbradford:
Fixes https://github.com/cfpb/design-system/issues/1398

## Changes

- Converts expandable header from a float to a flexbox layout.

## Testing

1. `yarn start` and view http://localhost:4000/design-system/components/expandables and resize the window and expand/collapse the expandables.

## Screenshots

Before:

Screen Shot 2022-10-11 at 12 31 57 PM

After:

Screen Shot 2022-10-11 at 12 31 01 PM

### DIFF
--- a/docs/pages/expandables.md
+++ b/docs/pages/expandables.md
@@ -19,10 +19,10 @@ variation_groups:
                       o-expandable__border">
               <button class="o-expandable_header o-expandable_target"
                       title="Expand content">
-                  <h3 class="h4 o-expandable_header-left o-expandable_label">
+                  <h3 class="h4 o-expandable_label">
                       Expandable Header
                   </h3>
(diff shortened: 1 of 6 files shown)


anselmbradford:
Note: I left the `o-expandable_header-right` for consistency, even though it doesn't have any styles applied any more. We could consider removing it entirely.

anselmbradford:
Edit: I have removed both `o-expandable_header-left` and `o-expandable_header-right`, as we really shouldn't have classes that designate positioning, and those two elements already have `o-expandable_label` and `o-expandable_link`, so we can target those instead.

flacoman91 (approved):
I viewed it on the demo site, and can confirm it's 🥇

PR #1414, opened 2022-10-12
Update cfpb-expandables.less to include `justify-content: space-between;`

anselmbradford:
This is needed for an expandable in the rural and underserved populations app.

## Changes

- Update cfpb-expandables.less to include `justify-content: space-between;`

## Testing

1. Check the expandables page in the PR preview and see that the expandables toggle is unchanged.


flacoman91 (approved):
👍  looks good

PR #1417, opened 2022-10-17
Set a min-width for expandable toggles

anselmbradford:
If an expandable has a very long header text, the toggle will wrap. This change sets a min-width on the toggle so it doesn't wrap, and also right aligns the text so the toggle aligns between the show and hide text.

## Changes

- Set a min-width for expandable toggles.

## Testing

1. Visit the PR preview and edit the header text for an expandable to a really long string and see how the toggle responds.

## Screenshots

Before:
Screen Shot 2022-10-17 at 12 24 57 PM

After:
Screen Shot 2022-10-17 at 12 25 03 PM

### DIFF
--- a/packages/cfpb-expandables/src/cfpb-expandables.less
+++ b/packages/cfpb-expandables/src/cfpb-expandables.less
@@ -101,6 +101,8 @@
   }
 
   &_link {
+    min-width: 60px;
+    text-align: right;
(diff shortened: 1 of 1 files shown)


flacoman91 (approved):
DPT approved.

Screen Shot 2022-10-17 at 12 31 04 PM

Issue #1423, opened 2022-10-18
Convert React components to Functional components


anselmbradford:
For a more modern React dev experience, we should update the class components to be functional components.

anselmbradford:
Some instructions https://stackoverflow.com/questions/69965343/convert-react-class-based-to-functional-component

sonnakim:
On hold till we move to Decap

anselmbradford:
https://github.com/cfpb/design-system/pull/1980 has been merged.

PR #1453, opened 2022-10-26
Convert two react class components to functional components

drob:
Refactors two react classes to use functional components instead of class components, per #1423.

## Changes

- Refactors `StringTrimmedControl` to use functional components instead of class components.

## Testing

1. PR checks should pass.
2. Any regressions should show up in the netlify preview.

## Notes

- I'm new to some of these libraries, so apologies if I'm doing something non-idiomatic. (Please let me know, if so!)
- There are three more class components in `admin/src/widgets`, which I'm planning on handling in a subsequent PR.

### DIFF
--- a/docs/admin/src/widgets/StringTrimmed.js
+++ b/docs/admin/src/widgets/StringTrimmed.js
@@ -9,11 +9,11 @@ export default function StringTrimmedControl({
   setActiveStyle,
   setInactiveStyle,
 }) {
-  const [state, setState] = React.useState({ value: value || '' });
+  const [inputValue, setInputValue] = React.useState(value || '');
 
   const handleChange = (event) => {
     onChange(event.target.value.trim());
-    setState({ value: event.target.value });
+    setInputValue(event.target.value);
   };
 
   return (
@@ -21,8 +21,8 @@ export default function StringTrimmedControl({
       type="text"
       id={forID}
       className={classNameWrapper}
-      value={state.value}
-      onChange={(event) => handleChange(event)}
+      value={inputValue}
+      onChange={handleChange}
       onFocus={setActiveStyle}
       onBlur={setInactiveStyle}
     />
@@ -32,7 +32,7 @@ export default function StringTrimmedControl({
 StringTrimmedControl.propTypes = {
   onChange: PropTypes.func.isRequired,
   forID: PropTypes.string,
-  value: PropTypes.node,
+  value: PropTypes.string,
   classNameWrapper: PropTypes.string.isRequired,
   setActiveStyle: PropTypes.func.isRequired,
   setInactiveStyle: PropTypes.func.isRequired,
--- a/docs/admin/src/widgets/StringTrimmed.js
+++ b/docs/admin/src/widgets/StringTrimmed.js
@@ -1,39 +1,39 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-export default class StringTrimmedControl extends React.Component {
-  static propTypes = {
-    onChange: PropTypes.func.isRequired,
-    forID: PropTypes.string,
-    value: PropTypes.node,
-    classNameWrapper: PropTypes.string.isRequired,
-    setActiveStyle: PropTypes.func.isRequired,
-    setInactiveStyle: PropTypes.func.isRequired,
+export default function StringTrimmedControl({
+  onChange,
+  forID,
+  value,
+  classNameWrapper,
+  setActiveStyle,
+  setInactiveStyle,
+}) {
+  const [state, setState] = React.useState({ value: value || '' });
+
+  const handleChange = (event) => {
+    onChange(event.target.value.trim());
+    setState({ value: event.target.value });
   };
 
-  state = {
-    value: this.props.value || '',
-  };
-
-  handleChange(event) {
-    this.props.onChange(event.target.value.trim());
-    this.setState({ value: event.target.value });
-  }
-
-  render() {
-    const { forID, classNameWrapper, setActiveStyle, setInactiveStyle } =
-      this.props;
-
-    return (
-      <input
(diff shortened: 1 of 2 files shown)


anselmbradford:
Thanks so much for the contribution @drob !

If I visit the deploy preview for this and attempt to edit a page, I'm getting 

Screen Shot 2022-10-26 at 10 06 55 AM

Do you see why this might be the case? Maybe a default value not set on the `useState` call?

P.S. if you have any interest in trying other conversions from class to functional components in a different context, https://github.com/cfpb/ccdb5-ui/ also needs this task done.

docs/admin/src/widgets/StringTrimmed.js
```diff
@@ -1,39 +1,39 @@
...
-    setInactiveStyle: PropTypes.func.isRequired,
-  };
+export default function StringTrimmedControl({
+  onChange,
+  forID,
+  value,
+  classNameWrapper,
+  setActiveStyle,
+  setInactiveStyle,
+}) {
```
flacoman91, line 11: this syntax should also be valid:

```
export default const StringTrimmedControl = ({
  onChange,
  forID,
  value = '',
  classNameWrapper,
  setActiveStyle,
  setInactiveStyle,
}) => {
```

docs/admin/src/widgets/StringTrimmed.js
```diff
@@ -1,39 +1,39 @@
...
+  const [state, setState] = React.useState({ value: value || '' });
```
flacoman91, line 12: Since this is an input, i would just make the value inside of useState always be a `string`

docs/admin/src/widgets/StringTrimmed.js
```diff
@@ -1,39 +1,39 @@
...
-      />
-    );
-  }
+  return (
+    <input
+      type="text"
+      id={forID}
+      className={classNameWrapper}
+      value={state.value}
+      onChange={(event) => handleChange(event)}
```
flacoman91, line 25: this can just be 
`onChange={handleChange}`

like how you have it down in StringWithInstructions

docs/admin/src/widgets/StringTrimmed.js
```diff
@@ -1,39 +1,39 @@
...
-        onBlur={setInactiveStyle}
-      />
-    );
-  }
+  return (
+    <input
+      type="text"
+      id={forID}
+      className={classNameWrapper}
+      value={state.value}
```
flacoman91, line 24: if you have the value changed from above from being an object, you don't need to make it state.value here.

docs/admin/src/widgets/StringTrimmed.js
```diff
@@ -1,39 +1,39 @@
...
+      onFocus={setActiveStyle}
+      onBlur={setInactiveStyle}
+    />
+  );
 }
+
+StringTrimmedControl.propTypes = {
+  onChange: PropTypes.func.isRequired,
+  forID: PropTypes.string,
+  value: PropTypes.node,
```
flacoman91, line 35: value should probably be a string here since that's going to be the default value that will go into the input.

docs/admin/src/widgets/StringTrimmed.js
flacoman91, line 12: If you're writing a lot of code that will use any of these react /redux hooks, you can also add this to the imports at the top

import React, { useState } from 'react';

so you can just do

const [foo, setFoo] = useState('');

drob:
@flacoman91 pushed a second commit that addresses your comments. Agreed these are all improvements. I was trying to keep the diff as minimal as possible – just class -> functional – and avoid doing additional refactoring while in there.

@anselmbradford I don't think I can reproduce that, since I'm not a repo contributor and so can't auth with the CMS. Thanks for the pointer to ccb5-ui, by the way! Let me know if you come across others.

docs/admin/src/widgets/StringTrimmed.js
```diff
@@ -1,39 +1,39 @@
...
+  classNameWrapper,
+  setActiveStyle,
+  setInactiveStyle,
+}) {
+  const [value, setValue] = React.useState(value || '');
 
-  state = {
-    value: this.props.value || '',
+  const handleChange = (event) => {
+    onChange(event.target.value.trim());
```
flacoman91, line 15: I think this onChange line should be deleted

and the next line can be:
setValue(event.target.value.trim())

flacoman91:
> @flacoman91 pushed a second commit that addresses your comments. Agreed these are all improvements. I was trying to keep the diff as minimal as possible – just class -> functional – and avoid doing additional refactoring while in there.
> 
> @anselmbradford I don't think I can reproduce that, since I'm not a repo contributor and so can't auth with the CMS. Thanks for the pointer to ccb5-ui, by the way! Let me know if you come across others.

Hi @drob!

Thanks for your contributions!  This is great stuff.  No worries.
Any help and feedback on our code is greatly appreciated.

I'm one of the maintainers of the [Consumer Complaint Database](https://www.consumerfinance.gov/data-research/consumer-complaints/search/).  If you ever wanted to take a look of the ins and outs of how that app works, you're more than welcome to check it out.  We're in the process of upgrading it to functional components, rewriting unit tests using react-testing-library, and also upgrading it to React 18.  I can also create issues with small tasks that anyone can pick up.

docs/admin/src/widgets/StringTrimmed.js
drob: Is that safe? Seems like it would change behavior vs the code that's there now, if an `onChange` function is passed in.

flacoman91: Sorry, you're right.

 I didn't see it was being passed in and was confused by the onChange inside of the input

PR #1497, opened 2022-11-22
Consolidate cf.gov responsive table styles, ignore print media

contolini:
Our responsive table styles are [duplicated in cf.gov](https://github.com/cfpb/consumerfinance.gov/blob/2ace1532d9b24595db6b436f008f02cc7e03ae19/cfgov/unprocessed/css/enhancements/tables.less#L125-L170). This PR consolidates them in the DS and a subsequent PR will remove them from cf.gov.

Also, printing a page triggers our smallest 600px breakpoint, causing our mobile styles to be applied. This is fine for most of our design patterns but it provides a degraded experience on pages with HTML tables. The responsive table media query has been updated to only apply to `screen` media types and not `print`.

See https://github.local/CFPB/1071-data-collection/issues/317#issuecomment-326594

## Changes

- Consolidated our responsive tables styles from cf.gov.
- Tables are not responsive with `print` media types.

## Testing

1. The tables page at the PR preview link should look identical to the [production tables page](https://cfpb.github.io/design-system/components/tables).
2. Try print previewing the above two pages and the responsive tables should no longer be stacked.

## Screenshots

Here's the results on our table-heavy [FIG page](https://secret-dev4-demo-sever.cfpb.gov/data-research/small-business-lending/filing-instructions-guide/202x-filing-instructions-guide/):

| before | after |
|--------|-------|
|  |  |

### DIFF
--- a/docs/pages/tables.md
+++ b/docs/pages/tables.md
@@ -95,7 +95,7 @@ variation_groups:
       structure, even on a smaller screen.
 
 
-      Note that tables are **not** responsive unless you add one of the small screen classes detailed below. Also note that the `data-label` attribute must be used to label each cell in a `table` for small screen responsive views.
+      Note that tables are **not** responsive unless you add one of the small screen classes detailed below. Also note that the `data-label` attribute must be used to label each cell in a `table` for small screen responsive views. Responsive layouts are **not** applied to tables when pages are printed.
     variations:
       - variation_name: Responsive stacked table
         variation_description: >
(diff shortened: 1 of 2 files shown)


anselmbradford:
🚀 🚀 __Just want to say this is a model PR. I ❤️‍🔥 the background context, testing steps, screenshots, etc.__ 🚀 🚀

wpears (approved):

PR #1498, opened 2022-11-22
Added right-to-left language support for Expandables component

brandon-rusk:
## Changes

- Small addition to the Expandable component's CSS to support right-to-left languages (such as Arabic)

## Screenshots

|Before|After|
|-|-|
|image|image|

## To-Do

- I suspect there needs to be a documentation addition in the Expandable component section to address it now supports right-to-left languages. Is that on me or the Design and Content team to add?


packages/cfpb-expandables/src/cfpb-expandables.less
```diff
@@ -209,3 +209,11 @@
     }
   } );
 }
+
+// Used when the set language reads right-to-left 
+[dir="rtl"] {
```
anselmbradford, line 214: Will this attribute appear on the top expandable element? Should we scope it to that? e.g.

```css
.o-expandable[dir="rtl"] {
```

(Or, perhaps better, move it up to between line 210-211 and make it `&[dir="rtl"] {`)

brandon-rusk: Hmmm, I'm not sure if I'm missing something but I've tested both those options and neither worked. Through all of my trial and error, I've only had success in changing `o-expandable_header` which is what the button is attached to. I am open to different solutions if we feel they would be better, but so far this is the best I've come up with

anselmbradford: `o-expandable_header` would stay. I'm just asking what element `dir="rtl"` appears on?

The full selector would be:

```css
.o-expandable[dir="rtl"] {
    .o-expandable_header {
    …
    }
}
```

Though you should be able to compact that to 

```css
.o-expandable[dir="rtl"] .o-expandable_header { 
   … 
}
```

brandon-rusk: Ah sorry, I misunderstood. `dir="rtl"` appears in the block `.block_-flush-top` which holds the expandable group within it. The problem is `dir="rtl"` is never applied to the `.o-expandable` element, which is why I put this code outside of it.

image

anselmbradford: Ah gotcha! Okay, looks like it's applied here to a `div` https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jinja2/v1/_includes/templates/render_block.html#L14

So I'd probably make it:

```css
div[dir="rtl"] .o-expandable_header { 
  …
}
```

That way it doesn't have to check every element in the dom tree for the `dir` attribute, only the `div`s (which will still be a large number, but not everything).

---

Zooming out a bit, will the pages this appears on always have a language setting on the `<html>` element? 
This [W3C page](https://www.w3.org/International/questions/qa-html-dir) says:
> If the overall document direction is right-to-left, add dir="rtl" to the html tag.
>
> Below the html tag, only use the dir attribute on structural elements on the rare occasions when the base direction needs to change in order for the text to display correctly.

If this will be always used on pages where the whole page is `rtl`, than that attribute should be added to the `html` element and you could do:

```css
:root[dir="rtl"] .o-expandable_header { 
  …
}
```

@contolini since it looks like you wrote that accessible block, any thoughts on one or the other of the above?

contolini: AFAIK we don't have a `dir` attribute on our root element because it reverses *all* content on the page and breaks the mega menu on smaller screens.

Screen Shot 2022-11-23 at 10 54 44 AM

Instead, we apply the attribute at the block level. This is not great and if we want to properly support RTL languages we should make our layouts direction-agnostic. Although before we do that, we should translate our header and footer content.

So with our current set-up, I think `div[dir="rtl"] .o-expandable_header {` is as specific as we can get to fix expandables.

mistergone: @brandon-rusk - I think what Ans suggests is a good change! 👍

brandon-rusk: Awesome, thanks for the feedback, all! I have updated the PR.

@anselmbradford, is there any docs that need to be updated in the DS to show Expandables now handle right-to-left languages?

anselmbradford: Yeah wouldn't be a bad idea to add an example of this. You can do that by adding a section in https://github.com/cfpb/design-system/blob/main/docs/pages/expandables.md (follow the basic expandable format, but wrap it in `<div dir="rtl">…</div>`.

mistergone (approved):

PR #1500, opened 2022-11-30
Update contribution release steps

anselmbradford:
Lerna has several bugs, such as not incrementing the changelog version number and the root package.json version number. This PR attempts to provide an updated hack-ish workaround set of steps to making a release.

## Changes

- Reduce gulp tasks down to only `gulp build`.
- Update contributing steps.
- Provide `pre-release` script to perform steps before lerna runs. 

## Testing

1. Run `yarn pre-release`.


wpears (approved):

anselmbradford:
Merging this for now, but this def needs revisiting…

PR #1505, opened 2022-12-07
Fix broken require to lunr in search

anselmbradford:
Oops this should be an import not a require in an esm world.

## Changes

- Change to an import for lunr search package.

## Testing

1. PR preview search should work.

### DIFF
--- a/packages/cfpb-icons/src/icons-svg-inline.js
+++ b/packages/cfpb-icons/src/icons-svg-inline.js
@@ -1,4 +1,4 @@
-const path = require('path');
+import path from 'path';
 
 /**
  * This file is a less plugin that gets included in a less file via the
--- a/docs/assets/js/search.js
+++ b/docs/assets/js/search.js
@@ -1,4 +1,4 @@
-const lunr = require('lunr');
+import lunr from 'lunr';
 
 /**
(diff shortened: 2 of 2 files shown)


contolini (approved):
The checks are failing but

ship it

packages/cfpb-icons/src/icons-svg-inline.js
```diff
@@ -1,4 +1,4 @@
-const path = require('path');
+import path from 'path';
```
chosak, line 1: @anselmbradford this change seems to have broken the GitHub Actions check that builds the docs and runs Lighthouse against them, see [here](https://github.com/cfpb/design-system/actions/runs/3643893956/jobs/6152560324):

> Cannot use import statement outside a module
>      Error in /home/runner/work/design-system/design-system/node_modules/@cfpb/cfpb-icons/src/icons-svg-inline.js (line null, column -1)

The command being run is `yarn build-netlify`. Would you mind taking a look at this failure?

anselmbradford: Spun out into https://github.com/cfpb/design-system/pull/1510 for easier review. Thanks!

PR #1510, opened 2022-12-08
Downgrade to commonjs syntax in less plugin

anselmbradford:
Fixes issue where https://github.com/cfpb/design-system/pull/1505 broke `yarn start`


chosak:
Thank you @anselmbradford, we are back in business: https://github.com/cfpb/design-system/actions/runs/3653150088

PR #1518, opened 2022-12-29
Invoke wdio config file with .cjs locally

wpears:
Currently building locally errors without this update


anselmbradford (approved):
This can go back to `.js` on the next version of wdio

PR #1515, opened 2022-12-27
Remove destroyInitFlag

wpears:
Highlights:
 - All table js has been removed. Table css associated with row-links and sortable tables has been removed.
 - dom-traverse.js has been removed, relying instead on browser native methods
 - the signature of multiselectutils' `create` method was simplified
 - Other minor removals

PR #1525, opened 2022-12-29
Remove combined button-groups

wpears:
Part of https://github.com/cfpb/design-system/pull/1515

### DIFF
--- a/packages/cfpb-buttons/src/molecules/button-groups.less
+++ b/packages/cfpb-buttons/src/molecules/button-groups.less
@@ -1,9 +1,5 @@
 @import (less) '../atoms/buttons.less';
 
-//
-// Button groups
-//
-
 .m-btn-group {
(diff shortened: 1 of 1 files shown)


packages/cfpb-buttons/src/molecules/button-groups.less
```diff
@@ -13,28 +9,3 @@
...
-
-  .a-btn {
-    border-radius: 0;
-    margin-right: 1px;
-  }
-
-  // TODO: There's a strange disappearing border issue on Chrome.
-  // This might have to be revisited.
-  .a-btn + .a-btn,
-  .a-btn__super + .a-btn__super {
```
anselmbradford, line 27: Interestingly, the super button is only used on 7 pages total on cfgov 🤔

anselmbradford (approved):

PR #1527, opened 2022-12-29
Remove destroyInitFlag and remove data-hook

wpears:
Part of https://github.com/cfpb/design-system/pull/1515


test/unit-test/src/cfpb-atomic-component/src/utilities/data-hook.spec.js
```diff
@@ -1,8 +1,9 @@
 import {
   add,
-  contains,
+  contains
 } from '../../../../../../packages/cfpb-atomic-component/src/utilities/data-hook.js';
```
anselmbradford, line 4: This can be `} from '../../../../../../packages/cfpb-atomic-component';`

anselmbradford (approved):
I waffled back-n-forth on this one as to whether this was useful for a complete API of these hooks, but I guess we can always bring this back if needed.

One minor path to update with more recent code changes in the spec, otherwise 👍

PR #1532, opened 2023-01-05
Adjust yarn commands

anselmbradford:
Ruby was setup for running the packages's unit tests, which wasn't necessary. This PR removes that redundant configuration and performs related changes.

## Changes

- Change `postinstall` script to `after-install` so that Ruby bundle command is not automatically run after `yarn install`, which is unneeded when working with the DS packages.
- Remove `run` commands from yarn.
- Break GHA workflow tests into unit tests and sauce workflows.
- Minor docs adjustments to reflect changes.

## Testing

1. PR checks should pass.


anselmbradford:
Merging in @wpears absence

PR #1543, opened 2023-01-12
Remove gulp

anselmbradford:
We had some issues with our published built assets:
 - While built assets (JS/CSS) were included in the published npm packages, only the built JS was included in this repo, the CSS being gitignored. It appears the JS used to be ignored when its suffix was `min.js`.
 - `cfpb-grid.css` is not included in its published npm package.
 - The published built CSS in the packages is not minified, soooo what's the point?

## Removals

- Remove gulp and gulp dependencies.
- Remove `scripts` from linting, since the JS utilities in that directory are removed in this PR.

## Changes

- Include build CSS in the packages.
- Include the 3rd-party license in `cfpb-design-system`.
- Add JS file suffixes in imports as is liked in ESM.

## Testing

1. Edit some JS and CSS in the packages and run `yarn build-packages` and see that the respective built files are changed.


contolini (approved):
🚀 🚀 🚀 🚀

PR #1550, opened 2023-01-17
Hide bullet on multiselect dropdown

anselmbradford:
@wpears [fixed this bullet on cfgov](https://github.com/cfpb/consumerfinance.gov/pull/7205), but I just noticed we have a duplicate `multiselect.less` in cfgov and the design-system and the edit never made it over here. 

## Changes

- Hide bullet on the multiselect.

## Testing

1. In the PR preview, visit the multiselect and add five items and there shouldn't have a bullet. The bullet looks like this:

Screen Shot 2023-01-17 at 1 43 41 PM

### DIFF
--- a/packages/cfpb-forms/src/organisms/multiselect.less
+++ b/packages/cfpb-forms/src/organisms/multiselect.less
@@ -99,6 +99,7 @@ select.o-multiselect {
   }
 
   &_options {
+    list-style-type: none;
     background-color: @white;
     padding: 0;
 


packages/cfpb-forms/src/organisms/multiselect.less
```diff
@@ -99,6 +99,7 @@ select.o-multiselect {
   }
 
   &_options {
+    list-style-type: none;
```
anselmbradford, line 102: Unlike https://github.com/cfpb/consumerfinance.gov/pull/7205, I put this on the `ul` at all times, just in case we ever have other messages in the future. If this seems like a bad idea, we can go with 7205.

anselmbradford:
Merging in @wpears absence

PR #1561, opened 2023-01-19
Remove AtomicComponent and update Expandables to use FlyoutMenu

anselmbradford:
Update Expandables to use FlyoutMenu, the same as Summaries and the Mega Menu in cfgov.

## Removals

- Remove AtomicComponent.js.

## Changes

- Update Expandables to use FlyoutMenu.

## Testing

1. PR preview should have expandables that work like https://cfpb.github.io/design-system/components/expandables


anselmbradford:
Gotta look at the performance on this, as I'm noticing this updated expandable is a little jankier in the animation.

contolini (approved):

wpears (changes requested):
One annoying complication of this very welcome change is that the TDP [expandable-facets](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/apps/teachers-digital-platform/js/expandable-facets.js) will also need to be updated to not use AtomicComponent.

Alternatively, you could just mirror the AtomicComponent code to the tdp app and just update the import path. That's probably what I would recommend.

wpears (approved):
I'm happy after https://github.com/cfpb/consumerfinance.gov/pull/7659

PR #1564, opened 2023-01-20
Transitions: Enforce having both expand and collapse transitions

anselmbradford:
If you have only a collapse transition, or an expand transition, then the transition completion events won't fire since there's no state to expand or collapse from.

## Additions

- Add `proxyEvents` method to BaseTransition to pass its events through to its child transitions.

## Changes

- Transitions: Enforce having both expand and collapse transitions.

## Testing

1. PR checks should pass.


anselmbradford:
Merging in @wpears absence

PR #1566, opened 2023-01-24
Refactor FlyoutMenu

anselmbradford:
## Changes

- Get rid of deferred collapse and halt transition in progress instead.
- Refactor FlyoutMenu to have states to make it easier to figure out what state it is in.
- Set MaxHeightTransition duration based one content height.
- Optimize some code with function guards, etc.
- Downcase event types.
- Add event log to transitions examples.

## Testing

1. Transitions page on PR preview should still work as before.


anselmbradford:
Merging in @wpears absence

PR #1594, opened 2023-02-24
Refactor email signup to simplify margins and handle different layouts

anselmbradford:
We can remove much of the margin settings in the email signup organism by using the `gap` property. Also, we gain no use of `m-btn-group`, other than introducing styles we need to override. So, we should create an element specific to the email signup, which I'll call `o-email-signup_buttons`.

## Changes

- Use `gap` property so that margin settings can be removed.
- Change `m-btn-group` to `o-email-signup_buttons`.

## Testing

1. Visit https://deploy-preview-1594--cfpb-design-system.netlify.app/design-system/patterns/e-mail-signup-forms and compare to https://cfpb.github.io/design-system/patterns/e-mail-signup-forms
There is a very slight difference in the gap between the button and link, but it's 1 or 2 pixels.

## Screenshots

Before:
Screen Shot 2023-02-24 at 3 58 59 PM

After:
Screen Shot 2023-02-24 at 3 59 20 PM

### DIFF
--- a/docs/pages/e-mail-signup-forms.md
+++ b/docs/pages/e-mail-signup-forms.md
@@ -29,7 +29,7 @@ variation_groups:
                       </label>
                       <input class="a-text-input a-text-input__full" id="email_2" name="email" type="email" placeholder="mail@example.com" required="">
                   </div>
-                  <div class="m-btn-group">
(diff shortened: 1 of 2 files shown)


brandon-rusk (approved):
Nice 🎉

PR #1595, opened 2023-02-24
Remove aria-expanded from content area

anselmbradford:
Lighthouse is reporting that the `aria-expanded` attribute on the content element of the flyout menu is not correct. 
The trigger should have an `aria-controls` or `aria-owns` property with the ID of the content area they control. However, this means we'd need to add a unique ID pair to each trigger and content. We could do this in the JS using something like https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID, but let's just strip `aria-expanded` off the content to begin with and see how we fair.

## Changes

- Remove aria-expanded from the content area.
- Refactor Summary component to originate the content fade from the button's `::after`.

## Testing

1. Should be no visual change.


anselmbradford:
Merging in @wpears absence.

PR #1597, opened 2023-03-01
Remove "expandables" variations

anselmbradford:
Removes whole "variations" section of the expandable. We should not follow this pattern of making arbitrary modifications to expandables. The "expanded" modifier is already covered in the page's second example. The "padded" modifier is covered in the first and subsequent examples on the page (this should probably be merged into the default expandable styles). Text and header elements are covered in the standard examples. `o-expandable_header-left` and `o-expandable_header-right` appear to be unused in the codebase and should be removed.

### DIFF
--- a/docs/pages/expandables.md
+++ b/docs/pages/expandables.md
@@ -144,15 +144,17 @@ variation_groups:
                   </p>
               </div>
           </div>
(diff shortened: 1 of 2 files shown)


contolini (approved):
I would like to apologize for the many ways I have hacked expandables across cf.gov.

PR #1608, opened 2023-03-09
cfpb-typography: Remove text-align: left from `u-block-link`

anselmbradford:
If we explicitly set a `text-align: left` on the jumplinks they won't align to the right by default on right-to-left (RTL) languages. By default in LTR languages it should be aligned left anyway.

## Changes

- cfpb-typography: Remove text-align: left from `u-block-link`.

## Testing

1. PR preview in https://cfpb.github.io/design-system/components/lists should be unchanged.


contolini (approved):
No idea why this would be needed. ✅

PR #1648, opened 2023-05-22
cfpb-core: Pulls in js-is-hidden utility from cf.gov

wpears:
We have a utility for showing elements when js is enabled (`.u-js-only`) but the opposite utility [lives in cf.gov](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/css/enhancements/utilities.less#L116-L121). This is silly!

I've brought the `.js-is-hidden` utility over here and renamed it to `.u-hide-if-js`, which seems clearer.

### DIFF
--- a/packages/cfpb-core/src/utilities.less
+++ b/packages/cfpb-core/src/utilities.less
@@ -18,7 +18,7 @@
 // And show it when JS is off
 //
 
-.js-is-hidden {
+.u-js-is-hidden {
   display: none !important;
   .no-js & {
     display: block !important;
--- a/packages/cfpb-core/src/utilities.less
+++ b/packages/cfpb-core/src/utilities.less
@@ -14,6 +14,18 @@
 }
 
 //
+// To hide an element when JS is on
+// And show it when JS is off
+//
+
+.js-is-hidden {
+  display: none !important;
+  .no-js & {
+    display: block !important;
+  }
+}
+
+//
 // Clearfix
 //
 
--- a/packages/cfpb-core/src/utilities.less
+++ b/packages/cfpb-core/src/utilities.less
@@ -18,7 +18,7 @@
 // And show it when JS is off
 //
 
-.u-js-is-hidden {
+.u-hide-if-js {
   display: none !important;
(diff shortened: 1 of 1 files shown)


packages/cfpb-core/src/utilities.less
```diff
@@ -13,6 +13,18 @@
   }
 }
 
+//
+// To hide an element when JS is on
+// And show it when JS is off
+//
+
+.js-is-hidden {
```
anselmbradford, line 21: I'm sorry I know it's more work, but could we update this to `u-js-is-hidden` to match the other utilities? Looks like it's only used in RAFT and CCDB, so these two references would just need an update:

- https://github.com/cfpb/consumerfinance.gov/blob/2c95562349cb5bf8273b1eda2229c19ba2116d9c/cfgov/jinja2/ccdb-complaint/ccdb-search.html#L22
- https://github.com/cfpb/consumerfinance.gov/blob/2c95562349cb5bf8273b1eda2229c19ba2116d9c/cfgov/v1/jinja2/v1/includes/blocks/raf_tool.html#L41

wpears: suuuuuure. It's not more work in this PR, just in the one that remove this class from cf.gov ^_^

anselmbradford (approved):
We could maybe use this in place of `m-notification__no-js` too:

- https://github.com/cfpb/consumerfinance.gov/blob/2c95562349cb5bf8273b1eda2229c19ba2116d9c/cfgov/jinja2/owning-a-home/_templates/no_js.html#LL3C1-L3C1
- https://github.com/cfpb/consumerfinance.gov/blob/2c95562349cb5bf8273b1eda2229c19ba2116d9c/cfgov/jinja2/rural-or-underserved/_templates/no_js.html#LL3C1-L3C1

The difference there is that RAFT and CCDB are react apps that just don't display when JS is off, whereas the other tools need to be manually hidden when JS is off, but the notification could use `u-hide-if-js` and not have the custom `__no-js` modifier.

PR #1650, opened 2023-05-23
Update deprecated Event.init to actual Event creation

anselmbradford:
Event.init is deprecated and actual events should be created instead.

## Changes

- Update deprecated Event.init to actual Event creation.

## Testing

1. `yarn jest` should pass. 
2. Selecting five items in the multiselect should show the max selections message.


wpears (approved):
Neat

Issue #1514, opened 2022-12-22
Icons showing "Promise" in preview


anselmbradford:
1. Visit https://cfpb.github.io/design-system/admin/#/collections/pages/entries/buttons
2. Scroll down and see this where there should be icons:

<img width="767" alt="Screen Shot 2022-12-22 at 4 41 26 PM" src="https://user-images.githubusercontent.com/704760/209231179-7f5d91dd-d5f9-4a2d-a43b-7cc3fb194d61.png">

sonnakim:
@anselmbradford has a PR open to move over to Decap, so this and other Netlify-specific bugs should stay on hold till that migration happens.

anselmbradford:
https://github.com/cfpb/design-system/pull/1980 has been merged

PR #1674, opened 2023-07-18
icons missing on buttons page

BrewCityBoy:
**Taken from issue #1514** 

1. Visit https://cfpb.github.io/design-system/admin/#/collections/pages/entries/buttons
2. Scroll down and see this where there should be icons:

Screen Shot 2022-12-22 at 4 41 26 PM


anselmbradford (approved):
Well that was easy.


Fixes https://github.com/cfpb/design-system/issues/1514

Issue #1698, opened 2023-08-09
[Bug] Max-width value is narrower than widest SVG icons


natalia-fitzgerald:
The SVG icon max-width value is set to 1em. This width is narrower than the widest SVGs which are 17px wide. This setting causes SVG icons that are wider than 16px to display smaller than intended. 

**To Reproduce**
1. Go to '[Design System - Iconography - Icons with text](https://cfpb.github.io/design-system/foundation/iconography#icons-with-text)'
2. Inspect round icon + Student loans 
3. Toggle the max-width (check and uncheck)
4. See size difference

**Expected behavior**
SVG round icons (or any SVG icon that is wider than 16px) next to paragraph text should load at the full size of the illustrator artboard and SVG icon viewBox. 17px (w) x 19px (h). 

**Screenshots**
| SVG with max-width of 1 em | SVG without max-width |  
|------------| ------------|
|![Screenshot 2023-08-09 at 10 31 52 AM](https://github.com/cfpb/design-system/assets/77348967/1ff451a2-3504-4bcb-86a4-41c4c2e8144b)|![Screenshot 2023-08-09 at 10 32 02 AM](https://github.com/cfpb/design-system/assets/77348967/0c739a0f-fa94-43f6-9ab0-a49c71a67531)|

Related issues/discussions: 
https://github.com/cfpb/capital-framework/issues/863
https://github.com/cfpb/design-system/issues/1675#issuecomment-1660596130 (and discussion)

PR #1699, opened 2023-08-09
Remove problematic icon max-width property

contolini:
We have dropped support for legacy IE and no longer need to set a max width for our icons. The property causes icons wider than 16px to display smaller than intended.

Fixes https://github.com/cfpb/design-system/issues/1698

## Removals

- Hard-coded SVG icon maximum width

## Testing

1. Follow the [reproduction steps](https://github.com/cfpb/design-system/issues/1698) with the [PR preview icon page](https://deploy-preview-1699--cfpb-design-system.netlify.app/design-system/foundation/iconography#icons-with-text). The icons should be 17px wide instead of 16px.

### DIFF
--- a/packages/cfpb-icons/src/cfpb-icons.less
+++ b/packages/cfpb-icons/src/cfpb-icons.less
@@ -43,9 +43,6 @@
   vertical-align: text-top;
   fill: currentcolor;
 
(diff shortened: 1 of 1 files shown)


natalia-fitzgerald (approved):
Looks great!

PR #1723, opened 2023-09-12
Swap Default "Value" text with Placeholder Text in Disabled Field

BrewCityBoy:
On the current [text inputs page](https://cfpb.github.io/design-system/components/text-inputs), the disabled input field shows the same value as the other input fields. As a user would not be able to interact with a disabled field, we should display the default placeholder text. 

| Current | After Update |
| -------- | ------------- |
|Screenshot 2023-09-12 at 6 31 06 PM|Screenshot 2023-09-12 at 6 29 47 PM|

### DIFF
--- a/docs/pages/text-inputs.md
+++ b/docs/pages/text-inputs.md
@@ -57,7 +57,6 @@ variation_groups:
                  type="text"
                  id="textinput-example-disabled"
                  placeholder="Placeholder text"
-                 value="Input text"
                  disabled>
         variation_specs: |-
           #### Default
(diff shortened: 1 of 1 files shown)


BrewCityBoy:
Per request, "value" text has been removed from default text input field
Screenshot 2023-09-13 at 10 51 40 AM

docs/pages/text-inputs.md
```diff
@@ -17,8 +17,7 @@ variation_groups:
           <input class="a-text-input"
                 type="text"
                 id="textinput-example-default"
-                placeholder="Placeholder text"
-                value="Input text">
+                placeholder="Placeholder text" >
```
anselmbradford, line 20: Very trivial, but the space could be removed before the closing `>`.

sonnakim (approved):

PR #1732, opened 2023-09-27
Convert `a-heading` inside `m-slug-header` to `m-slug-header_heading`

anselmbradford:
`m-slug-header` contained a heavily-customized `a-heading`. It seems like this should be an element of the slug header instead.

## Changes

- Convert `a-heading` inside `m-slug-header` to `m-slug-header_heading`

## Testing

1. The "slug heading" on https://deploy-preview-1732--cfpb-design-system.netlify.app/design-system/foundation/headings#slug-heading and https://deploy-preview-1732--cfpb-design-system.netlify.app/design-system/patterns/e-mail-signup-forms should be unchanged.

### DIFF
--- a/docs/pages/e-mail-signup-forms.md
+++ b/docs/pages/e-mail-signup-forms.md
@@ -7,7 +7,7 @@ variation_groups:
       - variation_code_snippet: >-
           <div class="o-email-signup">
               <header class="m-slug-header">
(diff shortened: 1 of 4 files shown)


natalia-fitzgerald (approved):
Looks good!

Issue #1730, opened 2023-09-26
"Meta-header" component not set to receive mobile styling for H4


natalia-fitzgerald:
**Describe the bug**
The "Meta-header" component is styled like an H4 but is not set to receive mobile styling for H4. Additionally, the component is called "Heading with icon" but seems to be more specific than that. It seems to specifically refer to either the "Meta-header" or a category within the meta-header? The heading for the component, the supporting copy, and the example bring up a lot of questions. I'd like to try to unpack this component so that I can update the documentation to better reflect exactly what's intended/referenced. 

Questions: 
- What specific component is being referenced in the Design System? 
  - Icon and H4
  - Icon and H4 plus pipe character and another icon and H4
  - Icon and H4 plus pipe character and another icon and H4 plus rule line and date

- What is considered the meta-header? 
  - Icon and H4 plus pipe character and another icon and H4 plus rule line and date?

## Design System content: 

![Screenshot 2023-09-26 at 3 34 02 PM](https://github.com/cfpb/design-system/assets/77348967/2fab0fdd-47ea-4bc4-8750-214ffa6145eb)
![Screenshot 2023-09-26 at 3 33 43 PM](https://github.com/cfpb/design-system/assets/77348967/0057aba1-e1a5-40dc-86ee-820076e60f9c)
![Screenshot 2023-09-26 at 3 33 47 PM](https://github.com/cfpb/design-system/assets/77348967/55163af5-959e-4fe7-a0d0-69dba6d28cc6)
![Screenshot 2023-09-26 at 3 33 53 PM](https://github.com/cfpb/design-system/assets/77348967/fdad09ee-4f45-48ec-ae87-15094e1a9cf0)

**To Reproduce**
1. Go to '[...](https://cfpb.github.io/design-system/foundation/headings#variations-1)'
2. Click on '[....](https://cfpb.github.io/design-system/foundation/headings#heading-with-icon)'
3. Read content for "Heading with icon" and read through HTML code snippet, implementation, and specs. 

**Expected behavior**
The component name, content, and example is specific and clear. And it receives the mobile styling. 

**Additional context**
Also appears: https://www.consumerfinance.gov/about-us/blog

@anselmbradford @jenn-franklin

natalia-fitzgerald:
Update: @anselmbradford and I are working on fixing this bug.

anselmbradford:
@natalia-fitzgerald Is this good to close?

PR #1757, opened 2023-09-29
Refactor m-meta-header

anselmbradford:
Refactors m-meta-header to handle different screen sizes better and to have a more uniform structure.

## Changes

- Refactor `m-meta-header`

## Testing

1. Check out the meta header in https://deploy-preview-1757--cfpb-design-system.netlify.app/design-system/foundation/headings#meta-header and resize to mobile

### DIFF
--- a/docs/pages/heading-hierarchy.md
+++ b/docs/pages/heading-hierarchy.md
@@ -16,9 +16,9 @@ variation_groups:
           Example of a non heading element:
           <p class="h1">A non-heading element</p>
           -->
-        variation_description: ""
-        variation_jinja_code_snippet: ""
-        variation_name: ""
+        variation_description: ''
+        variation_jinja_code_snippet: ''
+        variation_name: ''
         variation_specs: "| Heading   |
           Specs                                                            |
           Responsive behavior at < 601px |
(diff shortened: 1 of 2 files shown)


natalia-fitzgerald:
@anselmbradford 
The indent on the second line in “Office of Research Publication” (“Publication”) looks like it needs to move in slightly more in order to left align with the first line of text (“Office”).

IMG_9133

Also the horizontal gap between the icons and headings looks a little tight. Can we increase that gap slightly?

anselmbradford:
> The second line in “Office of Research Publication” (“Publication”) looks like it needs to move in slightly more in order to left align with the first line of text (“Office”).
 
 It is aligned, but because of the curve of the O and the straight edge of the P it looks a little off. Unfortunately, not much we can do here because we don't know what these beginning letters may be (and moving it would make an H and a P look misaligned, for example).
 
 > Also the horizontal gap between the icons and headings looks a little tight. Can we increase that gap slightly?
 
 Increase this to 5px.

Screenshot 2023-10-03 at 8 24 06 AM

anselmbradford:
Fixes https://github.com/cfpb/design-system/issues/1730

natalia-fitzgerald (approved):

PR #1791, opened 2023-10-25
Remove large pull quote

anselmbradford:
Accompanying PR to https://github.com/cfpb/consumerfinance.gov/pull/7999

I noticed the large pull quote is only used on https://www.consumerfinance.gov/data-research/consumer-complaints/

This pattern is very old and goes back to 2014 (added in https://github.com/cfpb/cf-typography/pull/17)

We have the "is large" option in Wagtail to support the one use case in the CCDB landing page.

Consulting with @natalia-fitzgerald, it was decided it would be fine for this page to have the regular pull quote, which I have updated on the live site.

## Removals

- Large pull quote.

## Testing

There wasn't an example of this in the DS, so there's nothing further to check beyond PR checks passing.


niqjohnson (approved):
Looks good. I was going to say we should also update our usage guidelines, but it turns out we don't have any usage guidelines on pull quotes, sooooo…

(We should probably write usage guidelines on pull quotes at some point, but that's definitely a separate issue.)

Issue #1728, opened 2023-09-22
Multiselect: On desktop drop-down options lose background fill 


natalia-fitzgerald:
**Describe the bug**
When options are shown on desktop and the user scrolls through the options, the background fill is lost after the user scrolls past the first 3 options. 

Related issue: 
https://github.com/cfpb/design-system/issues/1164

**To Reproduce**
1. Visit https://www.consumerfinance.gov/about-us/blog/
2. Click on a multiselect.
3. Type something if the options don't show up.
4. Scroll down in the options past the first 3 and see that the background behind the options scroll away.

**Expected behavior**
Background should say opaque behind the options.

**Screenshots**

<img width="277" alt="Screenshot 2023-09-22 at 6 21 25 PM" src="https://github.com/cfpb/design-system/assets/77348967/d472b3bf-28a9-43a7-ab12-f4859f380cd7">
<img width="268" alt="Screenshot 2023-09-22 at 6 21 31 PM" src="https://github.com/cfpb/design-system/assets/77348967/4018accb-04ad-420b-8182-d3762b0b7332">

anselmbradford:
I see this in Safari 16.6 (18615.3.12.11.2)

PR #1792, opened 2023-10-26
Multiselect: Remove `display: -webkit-box` property

anselmbradford:
https://github.com/cfpb/design-system/pull/1117 set `display: -webkit-box` to fix an issue with a Chrome v87. Since that time the underlying chromium bug has been closed, and it appears that this property is now causing a more serious issue in Safari. See https://github.com/cfpb/design-system/issues/1728

Fixes https://github.com/cfpb/design-system/issues/1728 

## Changes

- Remove `display: -webkit-box` property

## Testing

1. Check the PR preview in Safari and Chrome and see that the multiselect opens the same in both.

## Screenshots

Before: 

Screenshot 2023-10-25 at 8 15 29 PM

After: 

Screenshot 2023-10-25 at 8 15 45 PM

### DIFF
--- a/packages/cfpb-forms/src/organisms/multiselect.less
+++ b/packages/cfpb-forms/src/organisms/multiselect.less
@@ -78,12 +78,6 @@ select.o-multiselect {
     width: 100%;
 
     transition: max-height 0.25s ease-out;
-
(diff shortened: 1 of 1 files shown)


contolini (approved):
Buh-bye

PR #1801, opened 2023-11-03
Multiselect: Enable configuration at instantiation

meissadia:
These changes are in support of development for `design-system-react`.  In order to integrate the DS Multiselect into our React implementation, we need to be able to apply configuration for: 
- Max number of selections (number)
- Should display Tag/Pill (boolean)

## Additions

- feat: Support configuration options for Atomic components
- feat: [Mulitselect] Make rendering of selected Tags configurable
- feat: [Multiselect] Make maximum number of selections configurable

## Pending
- Should we support configuration per-Multiselect when calling `instantiateAll()`? Currently it takes a single configuration object but this could be updated to accept an array of configurations. 

## Testing

In parallel I'm integrating this component into the Design System React and these proposed changes are working great!

### DIFF
--- a/docs/pages/links.md
+++ b/docs/pages/links.md
@@ -24,14 +24,16 @@ variation_groups:
 
 
           <p>Here's the <a href="#">default link style</a>. For reference, here's the <a href="#" class="hover">hover link style</a>. Train your eyes on the <a href="#" class="focus">focused link style</a>. Jump to the <a href="#" class="active">active link style</a>. We’ve all been to the <a href="#" class="visited">visited link style</a>.</p>
-        variation_description: Inline links are regular weight and appear within the
+        variation_description:
+          Inline links are regular weight and appear within the
           text flow. They are regular weight and are used within paragraph of
           text.
         variation_name: Inline link
-        variation_implementation: "Links that appear in body text (`p`), link lists
+        variation_implementation:
+          'Links that appear in body text (`p`), link lists
           (`li`), or definitions (`dd`) are automatically underlined. To enable
           underlines elsewhere, e.g. to underline links in a `nav` element,
-          simply add a `border-bottom-width: 1px;` to the link."
+          simply add a `border-bottom-width: 1px;` to the link.'
         variation_specs: >-
           [Color variables for
           links](https://cfpb.github.io/design-system/development/variables#typography-1)
@@ -84,7 +86,8 @@ variation_groups:
                   <a class="m-list_link" href="#">List link 2</a>
               </li>
           </ul>
-        variation_description: List links (or call-to-action links) are standalone links
+        variation_description:
+          List links (or call-to-action links) are standalone links
           that highlight a users' next steps. They are medium weight and often
           used in unordered lists.
         variation_implementation: Call-to-action links are used in
@@ -92,7 +95,7 @@ variation_groups:
           unit
           groups](https://cfpb.github.io/design-system/patterns/info-unit-groups),
           among other places.
-        variation_jinja_code_snippet: ""
+        variation_jinja_code_snippet: ''
         variation_specs: >-
           * Avenir Next Medium, 16px
 
@@ -117,7 +120,8 @@ variation_groups:
           <a class="a-btn a-btn__link a-btn__warning" href="#">
                Destructive link
           </a>
-        variation_description: Destructive links provide a visual warning to users that
+        variation_description:
+          Destructive links provide a visual warning to users that
           clicking them will perform a destructive action, such as clearing
           entries in a form.
         variation_specs: |-
@@ -125,13 +129,14 @@ variation_groups:
           * Mid Dark Red (#c3381c)
       - variation_is_deprecated: false
         variation_name: Link with icon
-        variation_description: When used, an icon should appear after the text it
+        variation_description:
+          When used, an icon should appear after the text it
           represents. Each icon should be used exclusively and consistently for
           one action. The color and font-size of an icon should be the same as
           the text it represents, including state changes. Icons should never be
           underlined.
         variation_implementation: >-
-          
+
 
           * To prevent the link’s underline from extending under the icon, wrap the link text with a `span.icon-link_text`. There can be no whitespace between the text and the opening and closing span tags. Include the icon either prior to or after the `a-link_text`. It is important the text and icon are siblings to correctly handle underlines.
         variation_code_snippet: |-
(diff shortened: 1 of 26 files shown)


packages/cfpb-atomic-component/src/utilities/atomic-helpers.js
```diff
@@ -115,7 +116,7 @@ function instantiateAll(selector, Constructor, scope) {
     element = elements[i];
     if (contains(element, INIT_FLAG) === false) {
       inst = new Constructor(element);
-      inst.init();
+      inst.init(config);
```
anselmbradford, line 119: Thank you for this! This got me thinking about the architectural approach of configuring components at initialization.

On the one hand it's nice to have a call to `instantiateAll` always return the same thing, without any additional arbitrary config. However, we do need a way to specify different initial states of a component in some cases, such as the config you have in this PR. 

I looked at how we do this with Expandables. In Expandables, there is an option where an Expandable can be open on page load. For this, we add a CSS class [`o-expandable__onload-open`](https://github.com/cfpb/design-system/blob/fb976fcacecc7e5439b822e022be96befe1cbd6a/packages/cfpb-expandables/src/Expandable.js#L49). This class itself adds no styles directly (using JS to set additional CSS that sets the appearance). Per BEM naming, the `o-expandable` sets a certain appearance, and `o-expandable__onload-open` is a modifier on that base appearance. 

The trade offs of this are:

- Since the flag for this configuration appears in the markup, the site can be searched using the crawler for `o-expandable__onload-open` to find all expandables that are open on load. 
- This is possibly an abuse of BEM, since the modifier, `o-expandable_onload-open`, does not actually change the appearance directly itself.

Now, in regard to the multiselect, I could see the max selections being an attribute on the HTML, in the same way the native `<select>` element has a `size` attribute, for example. However, since the native HTML has a limited set of attributes, we would either need to add a custom `data-max-selections` attribute, or take a class-based approach, as we do with expandables. However, neither of these approaches are elegant. I think in the future, the multiselect will probably be a web component and we would set an attribute on the custom element.

So, in conclusion, I think this config approach on `instantiateAll` is fine, but two things should happen in the future through someone's effort:

- Expandables should align better with this same pattern.
- The configuration options should likely be set through some sort of attribute that appears in the markup.

packages/cfpb-atomic-component/src/utilities/atomic-helpers.js
```diff
@@ -102,10 +102,11 @@ function setInitFlag(element) {
  * @param {string} selector - Selector to search for in the document.
  * @param {Function} Constructor - A constructor function.
  * @param {HTMLElement} [scope] - A dom node in which to query the selector.
+ * @param {object} config - Configuration will be provided to the Constructor's init()
  *   If not supplied, it defaults to the `document`.
  * @returns {Array} List of instances that were instantiated.
  */
-function instantiateAll(selector, Constructor, scope) {
+function instantiateAll(selector, Constructor, scope, config = {}) {
```
anselmbradford, line 109: You'll want to update the associated unit tests for this change.

packages/cfpb-forms/src/organisms/Multiselect.js
```diff
@@ -27,6 +27,12 @@ const KEY_UP = 'ArrowUp';
 const KEY_DOWN = 'ArrowDown';
 const KEY_TAB = 'Tab';
 
+// Configuration default
+const DEFAULT_CONFIG = {
+  renderTags: true, // Allow the Multiselect to generate the Tag elements in the DOM
```
anselmbradford, line 32: In what case would we want this to be false? Is this to support https://www.consumerfinance.gov/data-research/prepaid-accounts/search-agreements/ ? Could we maybe change that to show tags with the multiselect?

meissadia: @anselmbradford This option was added to allow for a workaround (we render the Tags ourselves) to a bug we encountered while trying to integrate the DS Multiselect into the `design-system-react` implementation of the DS.  The DS icons are not rendering correctly. 
 
I've tried to debug but have yet to find a resolution. 

|DS Rendered Tags|React rendered Tags|
|---|---|
|Screenshot 2023-11-13 at 11 01 21 AM|Screenshot 2023-11-13 at 11 08 25 AM|

packages/cfpb-atomic-component/src/utilities/atomic-helpers.js
meissadia: Added unit tests for configuration support to both Multiselect and instantiateAll

test/unit-test/src/cfpb-atomic-component/src/utilities/atomic-helpers.spec.js
```diff
@@ -34,6 +35,12 @@ const HTML_SNIPPET = `
           </p>
       </div>
   </div>
+  <select name="test-select" id="test-select" class='o-multiselect' multiple>
```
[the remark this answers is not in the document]
meissadia: This element is within an automated unit test and not exposed to end users.

meissadia:
The failing browser test for Multiselect is due to a pop-up in the Sauce CI's browser and not due to any actual UI issue. 
 
I'm going to trigger a re-run and see if that gets things to pass, but I hope that false-negative doesn't block this PR. 
 
Video: https://app.saucelabs.com/tests/cb07f176fdbe4634a9d7b2a318b06eca?auth=cf927227628650267280588ab0fa1e63#80

packages/cfpb-forms/src/organisms/Multiselect.js
anselmbradford: Maybe add a `TODO: ` comment above this explaining why it's needed.

anselmbradford (approved):
Run `yarn lint` for good measure

meissadia:
> Run `yarn lint` for good measure

I've commited the changes made by `yarn lint`.  
 
I was not able to get the `yarn eslint ...` portion of the linting script to complete successfully.

meissadia:
Thank you very, very much for your guidance and feedback on this @anselmbradford!

docs/pages/links.md
```diff
@@ -198,16 +205,16 @@ variation_groups:
           appear in the following style and include both the original link text
           as well as a shortened URL.
         variation_code_snippet: Here's the <a href="#">link style</a> when printed.
-        variation_code_snippet_rendered: "<p>Here's the <a
-          href=\"https://consumerfinance.gov/about-us/blog\"
-          style=\"font-weight: 500;\">link style<span style=\"border-bottom: 1px
-          solid #ffffff; font-weight: 300;\">
-          (cfpb.gov/about-us/blog)</span></a> when printed.</p>"
+        variation_code_snippet_rendered: '<p>Here''s the <a
```
meissadia, line 208: These single quotes look iffy but everything is rendering as expected locally, so I'm not going to make any manual adjustments.

PR #1868, opened 2023-12-15
Normalize media query syntax

anselmbradford:
## Changes

- Normalize media query syntax by adding code comments for size and removing spaces inside parenthese.

## Testing

1. PR checks should pass.


wpears:
Sauce still seems troubled, thoughts?

anselmbradford:
> Sauce still seems troubled, thoughts?

Looking at swapping to Cypress.

PR #1871, opened 2023-12-21
Switch to Cypress

anselmbradford:
Switches the tests from Webdriver.io to Cypress to align with how our other repos have settled on Cypress. Note that the long timeouts are needed to wait for Netlify admin to load and do it's network calls.

## Changes

- Switch tests to Cypress

## Testing

1. PR checks should pass.
2. `yarn test:browser` should run headless Cypress locally.
3. Running Sauce Connect and a server locally, `yarn test:browser:cloud` should run the local tests on Sauce.


test/cypress/support/commands.js
wpears, line 1: We can probably remove this, no?

wpears:
I get `Too large read data is pending` on docs/netlify-cms.cy.js locally. I assume you haven't run into this?

wpears:
Oh reallllllly! Mine fails on `yarn test:browser`

wpears:
Oh I think my error is actually unrelated to that note, it's:
`(0 , marked__WEBPACK_IMPORTED_MODULE_3__.marked) is not a function`

Which afaict is an error in how `marked` is being imported..? Seems odd. In v4x there's no default import from `marked`, but our code isn't using default imports!

wpears (approved):

PR #1886, opened 2024-01-19
 Styling: Hover/Focus on TextInputs utilize dark colors

### LINKED ISSUE #1881: Text Input - Hover - Define state styling (Success/Warning/Error status)
As part of the React component library (and lending app) we are currently looking to nail down the state styling for text input (Error status).

### In the Design System we currently define the states as follows:
This is not visualized on the documentation page but is in the code. The hover state is Pacific while the other states are red. I am not sure whether this is intentional or a default style that was never customized for the error status. 
Screenshot 2024-01-17 at 2 09 26 PM

### If we apply the same visual logic that we do for the "[destructive link](https://cfpb.github.io/design-system/components/links#destructive-link)" we could go with a dark red on hover. 

That would look like this. What do you think? 
Screenshot 2024-01-17 at 3 15 54 PM

### I cross-referenced USWDS and they don't appear to have a separate hover style for their text inputs. 

**Default**
Screenshot 2024-01-17 at 4 11 50 PM

**Hover (No change)**
The only change here is a cursor change. 
Screenshot 2024-01-17 at 4 11 50 PM

**Focus**
Screenshot 2024-01-17 at 4 11 45 PM

@sonnakim for guidance

shindigira:
closes #1881 

## Changes
- Added new dark hover/focus colors to variables.md
- Added new dark hover/focus colors to text-inputs.md
- Updated text-inputs.md with the new dark hover/focus color stories

## How to Test 
- Test Link: https://deploy-preview-1886--cfpb-design-system.netlify.app/design-system/components/text-inputs
- **Code inspect** `text-inputs.less` and ensure no unintended styling changes affect other components
- Navigate to the `Variables` story section to view the new styling variables.
- **Navigate to the `TextInputs` story section and view the new dark hover/focus styles**. Hover and focus in the three normal status sections (success/error/warning)

## Screenshot(s)

Subtle dark colors -- only visible during hover (not focus)

|Before|After|
|---|---|
|Screenshot 2024-01-22 at 3 43 24 AM|Screenshot 2024-01-22 at 4 15 28 AM|


natalia-fitzgerald (changes requested):
@shindigira 
When I reviewed this with the design team we had only discussed changing the hover state to the darker color. Changing the focus state (or hover + focus) to the darker color was not the intent and would require further exploration/consideration. Perhaps as a first step we can change the hover to red (instead of blue) and put this additional change aside until I have time to discuss with the team next week? 

Or we can pause on this and focus on the checkbox Error status states as a first step.

packages/cfpb-forms/src/atoms/text-input.less
```diff
@@ -1,3 +1,79 @@
+/* Border and  Outline Status styling */
+.border-outline-error {
```
anselmbradford, line 2: For helper mixins like this we sometimes prefix them with `u-` to indicate that it's a one-off utility thing. I think that would help with these border helpers you have here. See a similar example usage in https://github.com/cfpb/design-system/blob/main/packages/cfpb-typography/src/atoms/links.less#L66

natalia-fitzgerald:
@shindigira 

I'd like to proceed as follows: 
- Default: Red 
- Hover: Dark Red
- Hover + Focus: Red
- Focus: Red

Thanks!

natalia-fitzgerald (approved):
Looks great! Nice work @shindigira.

anselmbradford:
I'll merge this after https://github.com/cfpb/design-system/pull/1973 goes in.

PR #1909, opened 2024-02-14
Streamline and update build

anselmbradford:
## Changes

- Replaces webpack with esbuild and removes webpack dependencies and config.
- Updates github actions to latest.
- Updates and fixes issues flagged by linter.
- Create `dist` directories in the published packages
- Fixes issue where cfpb-design-system package didn't have full JS file.

## Testing

1. `yarn build` should build the doc site's assets.
2. `yarn build:packages` should build the packages dist files.
3. `yarn build watch` should watch for changes in the doc site's assets.
4. `yarn start` should also watch the files.


wpears (commented):
Running `yarn build-packages` on a fresh checkout results in the following changes to /dist files:
Screenshot 2024-02-15 at 11 26 47 AM

This seems sub-optimal as it would clutter any given PR with these seemingly random /dist updates.

anselmbradford:
> This seems sub-optimal as it would clutter any given PR with these seemingly random /dist updates.

@wpears ahh thanks for catching this. I think this was the product of me doing a build on different branches and forgetting to commit the changes to this branch. I've committed the changes in https://github.com/cfpb/design-system/pull/1909/commits/3769d02ff464f96e299f2e03092fc4f2854722b4

The behavior going forward should be to only change the /dist/ files if their sources are changed.

wpears:
Hmmm, still getting:
Screenshot 2024-02-15 at 2 06 01 PM

anselmbradford:
> Hmmm, still getting:

Ah, that's because https://github.com/cfpb/design-system/pull/1911 went in, and then I updated the branch, but because 1911 only changed the source and not the dist files, the changes to the dist files from the inclusion of the range slider popped up here. Added in https://github.com/cfpb/design-system/pull/1909/commits/1cf9f9d7d485b169b44008b982e452031979bfc2

In the future, branches that change the source can run `yarn build-packages` and they'll have the dist files too, but in this case 1911 did not yet have that capability because this branch has not gone in yet.

wpears (approved):

PR #1918, opened 2024-02-27
Migrate Less color variables to CSS custom properties (variables)

anselmbradford:
## Changes

- Migrate Less color variables to CSS custom properties (variables).

## Testing

1. PR checks should pass.


wpears (approved):
Wow awesome! I wonder if, after this PR, we consider dropping some of the custom variables that are single use copies of the base color variables. Namely, things like `@table-scrolling-border` don't seem obviously helpful and actually I find them actively annoying as they break locality (I have to scroll to the definition [or use an IDE feature] to see what color is being set, rather than it just being a blindingly obvious `var(--gray-40)`)

anselmbradford:
> Wow awesome! I wonder if, after this PR, we consider dropping some of the custom variables that are single use copies of the base color variables. Namely, things like @table-scrolling-border don't seem obviously helpful and actually I find them actively annoying as they break locality (I have to scroll to the definition [or use an IDE feature] to see what color is being set, rather than it just being a blindingly obvious var(--gray-40))

Yes, I think that's a good idea. As far as I know these were added for theming, which we never really embraced.

PR #1919, opened 2024-02-27
Migrate custom BEM to getBEM syntax

anselmbradford:
## Changes

- Migrate custom BEM to getBEM syntax. The new syntax separates elements with a double underscore `__` and modifiers with a double dash `--`.

## Testing

1. PR checks should pass and the page preview should not be changed.


wpears (approved):
This is looking good! My main worry here is how breaking this change is for dependents, eg design-system-react and cfgov.

Does this warrant a major version bump? Otherwise, we need to make sure dependents are pinned to something before the next release (true for cfgov, not for d-s-r, until https://github.com/cfpb/design-system-react/pull/336 goes in).

contolini:
cc @billhimmelsbach @meissadia no rush but y'all will need to update DSR at some point

contolini (approved):
🚢

PR #1920, opened 2024-02-29
Merge normalize-css into repo

anselmbradford:
cfpb-core and cfpb-grid have a dependency on normalize-css, which hasn't been updated in ten years https://github.com/chrisdickinson/normalize-css

cfpb-grid has the dependency because we used to build an independent grid package on its own. 

We can't wholeheartedly remove this dependency without first checking that we haven't built upon the defaults it introduces. However, some of the more obvious obsolete ones I removed. You can see those in this commit https://github.com/cfpb/design-system/commit/8d767d5568f569e3764d68586c26960deb20ab03

## Removals

- Remove normalize-css dependency from cfpb-core and cfpb-grid.

## Changes

- Add hardcoded normalize.css to cfpb-core.

## Testing

1. PR preview should show no changes.


wpears:
Hmmmmm I wonder if we could do a very minimal reset like https://www.joshwcomeau.com/css/custom-css-reset/#our-finished-product-10 and then visual regression test it on our top X pages. @contolini how's your tool coming along?

anselmbradford:
> Hmmmmm I wonder if we could do a very minimal reset like https://www.joshwcomeau.com/css/custom-css-reset/#our-finished-product-10 and then visual regression test it on our top X pages. @contolini how's your tool coming along?

Simplifying this further makes sense. In the interim we could supply this PR's `normalize.css` and use that to whittle down to something slimmer in the future. The aim of this PR was mostly to remove the external dependency.

contolini:
> @contolini how's your tool coming along?

The brute force viz tester? gtg https://github.com/cfpb/cfgov-visual-regression?tab=readme-ov-file#typical-workflow

wpears (approved):

PR #1921, opened 2024-03-01
Remove mix(…) utility and replace with hardcoded palette values

anselmbradford:
We use this `mix(…)` Less utility to mix the background and text colors in buttons that have dividing lines. This creates a color that is almost, but isn't exactly, in our color palette. This technique is helpful if we were to theme the buttons and set the background and text color and have the divider automatically match. However, `mix(…)` being a Less utility means it will only take Less variables for the colors and won't work with CSS properties (variables). In the future, if we migrate to Sass and want buttons to be themeable, we could re-introduce this technique using the Sass helper of the same name (which may be more flexible), but in the mean time it seems easiest to set a variable for the divider color to colors from our palette. 

This appears on cfgov in the pagination buttons, the RSS feed button, and the print button.

## Changes

- Remove mix(…) utility and replace with hardcoded palette values.

## Testing

1. The buttons preview page will show a slightly different divider color, but the difference should be almost indiscernible.


anselmbradford:
Actually we could do this natively with https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix, but probably best to just pick a color from our palette? I could go either way.

wpears:
Another option is just to not set the border color and let it use the text color. It makes the dividing line a little more prominent, so perhaps this is worse and not worth the simplicity?

wpears (approved):

PR #1938, opened 2024-03-26
Remove unused `@text-shadow` argument

anselmbradford:
This argument for `heading-5(…)` and `heading-6(…)` does not to be anything more than an argument. There doesn't appear to be an implementation.

## Removals

- Remove unused `@text-shadow` argument

## Testing

1. PR checks should pass.


contolini (approved):
🎉

Issue #1943, opened 2024-04-02
Bug: Error alert link text fails accessibility contrast test


chosak:
The error alert style fails accessibility contrast checks. See https://cfpb.github.io/design-system/components/alerts#error:

<img width="794" alt="image" src="https://github.com/cfpb/design-system/assets/654645/d741a37e-a6d8-4953-85bd-fb55656c7389">

From a Chrome Lighthouse audit:

<img width="718" alt="image" src="https://github.com/cfpb/design-system/assets/654645/d08d9398-6459-430a-a1fe-00f6cb064e93">

Background color: #fbefec
Foreground color: #0072ce
Ratio: 4.34:1, which fails WCAG AA "Normal text" check, according to the [WebAIM checker](https://webaim.org/resources/contrastchecker/):

<img width="728" alt="image" src="https://github.com/cfpb/design-system/assets/654645/607f91ff-cc7e-43e5-a0b3-2aa67ba2aa6a">

Ratio needs to be at least 4.5:1.

@sonnakim for awareness

PR #1944, opened 2024-04-02
cfpb-notification: Add color contrast safe colors for links in an error notification

anselmbradford:
Fixes https://github.com/cfpb/design-system/issues/1943

The standard link color on a red background in the alert notifications is not accessible.

## Changes

- cfpb-notification: Add color contrast safe colors for links in an error notification

## Testing

1. PR checks should pass. Visit the alert page in the PR preview and run Google Lighthouse and it should pass the accessibility check with 100%.


chosak (approved):
Thank you! For posterity we should note that this PR closes #1943.

PR #1950, opened 2024-04-04
fix: o-text-introduction - Limit content width to 41.875rem

meissadia:


natalia-fitzgerald (approved):
Looks good!

packages/cfpb-layout/src/organisms/text-introduction.less
```diff
@@ -6,5 +6,6 @@
   & > * {
     margin-top: 0;
     margin-bottom: 15px;
+    max-width: 670px;
```
anselmbradford, line 9: This should be `max-width: 41.875rem`, which is how we set it in several other places. Aside—we should consolidate these max-width settings some time…

anselmbradford (changes requested):

packages/cfpb-layout/src/organisms/text-introduction.less
meissadia: Updated, thanks for catching that.

anselmbradford (approved):

PR #1958, opened 2024-04-18
Fix search

anselmbradford:
Search broke at some point because embedded SVG icons were messing up the quoting of the search string. 

## Changes

- Properly escape search string data.
- Add cypress test for search.
- Left align search results and cfpb logo.

## Testing

1. http://localhost:4000/design-system/search/ should work for searching.


wpears (approved):
Thanks!

PR #1964, opened 2024-04-24
cfpb-typography: Remove `a-heading__icon` class

anselmbradford:
I can't find any usage of `a-heading__icon`, and it's not in our docs either (outside of the variable reference).

## Changes

- Remove `a-heading__icon` class and associated variables and documentation.

## Testing

1. PR checks should pass.


anselmbradford:
I think this was refactored away in https://github.com/cfpb/consumerfinance.gov/pull/7975

contolini (approved):
🔪

PR #1987, opened 2024-05-29
Remove `a-link--icon-before-text` and `a-link--icon-after-text`

anselmbradford:
Jump links required both the `a-link--jump` and `a-link--icon-after-text` (or `a-link--icon-before-text`) modifiers to correctly layout with an icon. However, using flexbox we can have these links correctly layout whether or not they have an icon and whether or not it's on the left or right-hand side.

## Removals

- Remove `a-link--icon-before-text` and `a-link--icon-after-text`

## Changes

- Convert Jump Links to use flexbox.

## Testing

1. Visit https://deploy-preview-1987--cfpb-design-system.netlify.app/design-system/components/links and check out the jump links across screen sizes.

## Screenshots

Screenshot 2024-05-29 at 6 01 40 PM

Screenshot 2024-05-29 at 6 01 47 PM

### DIFF
--- a/docs/pages/featured-content-modules.md
+++ b/docs/pages/featured-content-modules.md
@@ -79,8 +79,7 @@ variation_groups:
                       commodo cu.
                   </p>
                   <a class="a-link
-                            a-link--jump
-                            a-link--icon-after-text">
+                            a-link--jump">
                       <span class="a-link__text">Read more about the feature</span>
                   </a>
               </div>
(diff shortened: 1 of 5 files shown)


wpears (approved):
Cool. Love these little changes that shave off the cruft

anselmbradford:
I realized this won't work out if we have a jump link that has an external icon, but I'll merge this and then open a follow up PR that handles that case and does some other cleanup.

PR #1989, opened 2024-05-30
Remove `a-link--no-wrap`, `a-link-icon`, and `m-list__link`

anselmbradford:
## Removals

- Remove `a-link--no-wrap`, `a-link-icon`, and `m-list__link`

## Changes

- Convert list links to directly use jump links.

## Testing

1. Visit https://deploy-preview-1989--cfpb-design-system.netlify.app/design-system/components/links#list-link and see that the list of jump links looks correct across screen sizes.

## Screenshots

Screenshot 2024-05-30 at 12 23 29 PM

Screenshot 2024-05-30 at 12 23 50 PM


wpears (approved):
Cool. I wonder if "jump links" are still a good name for this pattern. Makes me think there's anchor-linking behavior built in, but that's not really what it's about.

anselmbradford:
> Cool. I wonder if "jump links" are still a good name for this pattern. Makes me think there's anchor-linking behavior built in, but that's not really what it's about.

It's a good question!

PR #2009, opened 2024-08-02
Remove `a-legend` class | Move legends inside fieldsets

anselmbradford:
The `a-legend` class was just an h4 with additional IE legacy CSS. With the legacy CSS removed it just becomes an h4, so best to just set `<legend class="h4">` instead.

## Removals

- Remove `a-legend` class

## Changes

- Move legends inside fieldsets

## Testing

1. Check the PR preview legends page and it should be unchanged from production.


contolini (approved):
I support simplicity

PR #2026, opened 2024-08-22
Tags: Add initial tag styling

anselmbradford:
## Additions

- Tags: Add initial tag styling

## Testing

1. Check the tags page at https://deploy-preview-2026--cfpb-design-system.netlify.app/design-system/components/tags

### DIFF
--- a/docs/pages/tags.md
+++ b/docs/pages/tags.md
@@ -11,7 +11,8 @@ variation_groups:
   - variations:
       - variation_is_deprecated: false
         variation_name: Topic tag
-        variation_description: Topic tags appear with content such as blog posts and
+        variation_description:
+          Topic tags appear with content such as blog posts and
           press releases to display associated topics. When clicked, a topic tag
           takes a user to a filtered view of other items tagged with the same
           topic tag. In this filtered view, a topic tag appears as a filter tag.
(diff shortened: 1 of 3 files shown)


jenn-franklin (changes requested):
One small edit: In looking at filter tags on the [blog](https://www.consumerfinance.gov/about-us/blog/?topics=access-to-credit) and [Selects page](https://cfpb.github.io/design-system/components/selects#multiselect), it looks like the text should be regular vs. medium weight.

anselmbradford:
> One small edit: In looking at filter tags on the [blog](https://www.consumerfinance.gov/about-us/blog/?topics=access-to-credit) and [Selects page](https://cfpb.github.io/design-system/components/selects#multiselect), it looks like the text should be regular vs. medium weight.

@jenn-franklin 
So in TDP and prepaid it is medium (also CCDB, but that's its own beast) 

https://www.consumerfinance.gov/consumer-tools/educator-tools/youth-financial-education/teach/activities/?grade_level=3
https://www.consumerfinance.gov/data-research/prepaid-accounts/search-agreements/?prepaid_type=GPR+%28General+Purpose+Reloadable%29

Do we want all cases to be regular or medium? Or do we want two versions (which I'd discourage).

jenn-franklin:
Omg 😩 

I agree with you regarding discouraging two versions. I think we should go with regular weight since this is currently our standard in places like the blog, DS Selects page, and Figma file. 

I did contemplate if medium might look better and the fact that medium would be more in line with the button's styling. But we want to ensure tags aren't confused with buttons, so I think regular weight is actually better in that regard. And since regular weight is accessible on this teal background, we're good from a contrast perspective.

jenn-franklin (approved):

PR #2027, opened 2024-08-22
Remove `$font-size` variable

anselmbradford:
The heading mixins allow passing in a font-size override that is not used in the wild. There is also a `$font-size` variable that abstracts the `$size-*` variables. 

## Changes

- Remove `$fs` argument from `heading-*` mixins.
- Remove `$font-size` variable.

## Testing

1. The heading examples should not be changed.

https://deploy-preview-2027--cfpb-design-system.netlify.app/design-system/foundation/headings

vs

https://cfpb.github.io/design-system/foundation/headings


contolini (approved):
🗑️

PR #2031, opened 2024-09-04
Add topic tags | Re-arrange links page | Adjust sass imports

anselmbradford:
## Changes

- Add topic tags
- Re-arrange links page so that jump links outside of a group aren't placed adjacent to each other.
- Adjust sass imports to absolute paths.
- Fix grid mixin division deprecation warning.

## Testing

1. Check the topic tags across screen sizes on https://deploy-preview-2031--cfpb-design-system.netlify.app/design-system/components/tags
2. See link list moved below jump links and jump link examples reduced to single links https://deploy-preview-2031--cfpb-design-system.netlify.app/design-system/components/links


jenn-franklin:
Awesome. A few style tweaks:

- **Text color:** Should be #5A5D61 (not #676A6F)
- **Default underline:** Should be 1px, dotted, #FFB858. I'm struggling to inspect margins in the code. I _think_ the underline should sit 5px below the text. 
- **Hover underline:** Make it the same distance below the text as the default underline, which I _think_ should be 5px. Everything else (color, solid line) looks good to me as is.
- **Bullet:** Should be bigger and also more margin between it and the text. Looks like it's set at a size of 12px right now, and when looking at the Wayback machine, I see it's 17px, though the code setup is a little different. Also, I'm struggling to inspect the margin but I _think_ it's 17px? Comparison screenshots re: text size:

| Wayback machine    | DS preview |
| -------- | ------- |
| Screenshot 2024-09-04 at 1 58 12 PM  | Screenshot 2024-09-04 at 1 58 38 PM  |

jenn-franklin (approved):

jenn-franklin (approved):

contolini (commented):
Couple small questions but looks good 🚀

PR #2054, opened 2024-09-18
Add new tooltip component to DS

contolini:
We have two cf.gov apps that use tooltips but limited documentation on best practices or how to use them in new projects. This PR adds one implementation to the DS. It requires a third party library called [Tippy.js](https://atomiks.github.io/tippyjs/).

It requires code in the following format:

```
Some sentence that requires additional clarity <span data-tooltip="relevant-tooltip">{% include 'icons/help-round.svg' %}</span>
<template id="relevant-tooltip">
    <div class="tippy-heading">Here's the tooltip heading</div>
    <div class="tippy-body">And the tooltip body</div>
</template>
```

Creating the CFPB theme requires augmenting the [`tippy-* classes`](https://atomiks.github.io/tippyjs/v6/themes/#creating-a-theme). There's no CSS that's usable without the Tippy library so I didn't create an `o-tooltip` organism.

## Testing

1. Visit the [tooltip page](https://deploy-preview-2054--cfpb-design-system.netlify.app/design-system/components/tooltips) at the PR preview website.

## Screenshots

tooltip

## Notes

- It doesn't feel great adding Tippy.js to the main package.json when its code will be used on few pages.

## Todos

- Add more documentation about usage and design guidelines.

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


jenn-franklin:
Amazing. 

Proposing some tweaks to the content:

**Intro**
A tooltip provides short, descriptive information when a user hovers or focuses on an element. It contains helpful but non-critical information and is useful in a space-constrained user interface. 

**Guidelines**
Be judicious in using tooltips and explore other design options that keep content visible before choosing to use a tooltip. Only consider using a tooltip for short, non-critical information in a space-constrained user interface. Because a tooltip is hidden until activated, ensure content within a tooltip is not essential for completing a task on the page. For more guidance, see the [USWDS tooltip page](https://designsystem.digital.gov/components/tooltip/). 

**Behavior**
- Activate a tooltip by hovering or focusing on a help (question mark) icon situated next to text.
- Press the escape key to dismiss open tooltips. 
- When a tooltip is at the edge of the user’s viewport, it should automatically reorient itself away from the edge of the screen to prevent content clipping.

**Accessibility**
As USWDS states, tooltips are progressive enhancements for the `title` attribute and will display as the `title` attribute if the component doesn’t initialize. When testing tooltips for accessibility, ensure they are compliant with [USWDS tooltip accessibility tests](https://designsystem.digital.gov/components/tooltip/accessibility-tests/).

jenn-franklin (changes requested):
Amazing. 

Proposing some tweaks to the content:

**Intro**
A tooltip provides short, descriptive information when a user hovers or focuses on an element. It contains helpful but non-critical information and is useful in a space-constrained user interface. 

**Guidelines**
Be judicious in using tooltips and explore other design options that keep content visible before choosing to use a tooltip. Only consider using a tooltip for short, non-critical information in a space-constrained user interface. Because a tooltip is hidden until activated, ensure content within a tooltip is not essential for completing a task on the page. For more guidance, see the [USWDS tooltip page](https://designsystem.digital.gov/components/tooltip/). 

**Behavior**
- Activate a tooltip by hovering or focusing on a help (question mark) icon situated next to text.
- Press the escape key to dismiss open tooltips. 
- When a tooltip is at the edge of the user’s viewport, it should automatically reorient itself away from the edge of the screen to prevent content clipping.

**Accessibility**
As USWDS states, tooltips are progressive enhancements for the `title` attribute and will display as the `title` attribute if the component doesn’t initialize. When testing tooltips for accessibility, ensure they are compliant with [USWDS tooltip accessibility tests](https://designsystem.digital.gov/components/tooltip/accessibility-tests/).

jenn-franklin (approved):

contolini:
@anselmbradford after trying a handful of techniques to conditionally load the tooltip third-party dependency only when a tooltip is used, I ultimately decided to not bundle the tooltip code with the rest of the DS and instead added it as a [separate entrypoint](https://github.com/cfpb/design-system/pull/2054/files#diff-3b081710c2f17a625393a0c3581573c62922efe2fa14b5519cec7afd17fde398R5) using the [`exports` field](https://nodejs.org/api/packages.html#package-entry-points). This allows developers to selectively add tooltips to their project by importing [`@cfpb/cfpb-design-system/tooltips`](https://github.com/cfpb/design-system/pull/2054/files#diff-db0409ac0cd51b754f0b79dfa9dffb02ac20fe1cca9e2f31b0249137f11d8aa5R14). I added a note to the variation code snippet with usage instructions.

packages/cfpb-design-system/src/components/cfpb-tooltips/tooltip.scss
```diff
@@ -0,0 +1,27 @@
+@use 'sass:math';
+@use '@cfpb/cfpb-design-system/src/abstracts' as *;
+@import 'tippy.js/dist/tippy.css';
+@import 'tippy.js/dist/border.css';
```
anselmbradford, line 4: Does it mess things up to use `@use` here?

packages/cfpb-design-system/src/components/cfpb-tooltips/index.js
```diff
@@ -0,0 +1,79 @@
...
+              if (event.key === 'Escape') {
+                hide();
+              }
+            }
+            return {
+              onShow() {
+                document.addEventListener('keydown', onKeyDown);
+              },
+              onHide() {
+                document.removeEventListener('keydown', onKeyDown);
```
anselmbradford, line 60: Does it make any sense to target the body with these events. `document.body` instead of `document`?

anselmbradford:
When resolving conflicts, you'll want to delete the `npm-packages-offline-cache` folder.

anselmbradford:
Should we have a cursor change on hover?

anselmbradford:
If JS is off, it seems like the (?) icons should be hidden.

packages/cfpb-design-system/src/components/cfpb-tooltips/index.js
contolini: I don't think it really matters. The event will bubble up to both of them. I suppose there could be a situation where the page's `body` isn't filling the entire viewport, in which case pressing escape wouldn't be captured. Although I think the spec says that if no element is focused, it's fired on the `body`. So for clarity I'll just change it to `document.body`.

contolini:
> Should we have a cursor change on hover?

I say no because a `pointer` cursor indicates something will happen on click (and nothing does in our case).

> If JS is off, it seems like the (?) icons should be hidden.

I originally didn't do this because I figured there are scenarios where a user might want some words in the middle of a sentence to trigger a tooltip and it'd be bad if those words disappeared when JS is disabled. But our guidance explicitly states that help icons should trigger them so I'll add this feature and include a note about it in the guidance section.

packages/cfpb-design-system/src/components/cfpb-tooltips/tooltip.scss
contolini: Nah, great idea.

anselmbradford:
> I say no because a pointer cursor indicates something will happen on click (and nothing does in our case).

If you're hovered on the (?), the tooltip shows. If you then click the (?), the tooltip is dismissed—so that's the action I was thinking the cursor change would indicate. 

If not a cursor change, should there be a hover change on roll over? Or is that too button like? I noticed https://designsystem.digital.gov/components/tooltip/ has both a hover color change and a cursor change, but they also show the tooltip on buttons, so I'm unclear if that's completely covered by the buttons themselves, or if some of that is for the tooltip. Maybe @jenn-franklin can tell us how it should be and that's that.

jenn-franklin:
:+1: to a cursor change + click to close. No hover change. Thanks for calling this out!

anselmbradford:
Just in case you missed it—`npm-packages-offline-cache` still needs to be deleted.

packages/cfpb-design-system/package.json
```diff
@@ -21,5 +24,19 @@
...
+    "auto-changelog": "2.5.0"
+  },
+  "auto-changelog": {
+    "commitLimit": false,
+    "ignoreCommitPattern": "^Update \".+\" page",
+    "output": "./CHANGELOG.md",
+    "startingVersion": "v3",
+    "template": "./changelog-template.hbs",
+    "unreleased": true
+  }
```
anselmbradford, line 41: This should be removed

anselmbradford (approved):

PR #2055, opened 2024-09-20
Refactor heading mixins

anselmbradford:
We have widely used heading mixins for setting heading sizes. The `heading-1`-`heading-6` mixins set font and line height sizing, and add bottom margin. The `h1`-`h6` mixins extend the `heading-` mixins and add top margin, and for `h4` and up, add responsive styling for mobile sizes.

The heading mixins are either used as the standalone headings, or as the extended responsive version. In several uses, the margin of the heading is overridden. This is done by calling the mixin and then adding a separate margin property below the mixin to override the mixin. Additionally, rarely the text weight also gets overridden (e.g. on the h2 on the homepage 50-50 hero).

Sass has deprecated placing declarations below mixins (see https://sass-lang.com/documentation/breaking-changes/mixed-decls/), to align with the direction the CSS standard is going. We can still do this by wrapping the declaration in `& { … }`, but ultimately this generates a redundant selector block, so it's probably best to refactor things so that the mixins can be placed last. 

This PR merges the `h1`-`h6` mixins into the `heading-1`-`heading-6` mixins so there is only one set of heading mixins to be dealing with. It makes these mixins take several arguments to remove the top or bottom margin, specify if its responsive or not (if it's heading-4 or larger), and whether the font-weight can be adjusted (if it's `heading-2`-`heading-4`—this is based on real-world usage of this override). 

## Changes

- Remove `h1`-`h6` mixins and merge them into the `heading-1`-`heading-6` mixins.
- Refactor the heading mixins to make the bottom and top margin optional, make the responsive styling optional, and make the font-weight customizable. 

## Testing

1. Check the changed components on their respective PR preview pages.

At minimum `@include h1`-`@include h6` should map to `@include heading-1`-`@include heading-6`. 

`@include heading-1`-`@include heading-4`, should map to `@include heading-1($is-responsive: false)`-`@include heading-4($is-responsive: false)`.

Heading 5 and 6 don't have responsive styles, so `@include heading-5`/`@include heading-6` should be unchanged. 

## Notes

- I don't love this new API for the heading mixins, but couldn't figure out something better. I don't love that each mixin doesn't have all the same arguments, but that's how these are used in the wild, so I guess that's how it needs to be.


packages/cfpb-design-system/src/components/cfpb-forms/label.scss
```diff
@@ -18,17 +18,11 @@
...
-    /* stylelint-disable no-duplicate-selectors */
-    & {
-      display: block;
+    margin-bottom: math.div(10px, $size-iv) + em;
 
-      // Overwrites heading-4 margin.
-      margin-bottom: math.div(10px, $size-iv) + em;
-    }
-    /* stylelint-enable */
+    @include heading-4($has-margin-bottom: false);
```
contolini, line 25: Why not combine these two lines by changing the mixin API to something like `heading-4($margin: <whatever>);`? And give `$margin` a default value in the mixin definition.

contolini (approved):
A solid solution 👍

packages/cfpb-design-system/src/components/cfpb-forms/label.scss
anselmbradford: Mostly because `margin-top` has a different value across breakpoints in `heading-1` and `heading-2`. In the FIG these get overridden as a flat value across all breakpoints ([i.e.](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/apps/filing-instruction-guide/css/main.scss#L14)), whereas in other cases it adjusts across breakpoints. So I didn't see a non-confusing way to handle both of those scenarios. Margin-bottom could be `$margin`, as could the margin on `heading-3` through `heading-6`, but it seemed confusing to have a different margin signature based on the margin and heading size. I don't love this setup though, so certainly open to better ways to engineer this. Maybe this can go in and it can get refined from here.

PR #2058, opened 2024-09-25
Swap lerna with release-it

anselmbradford:
## Changes

- Swap lerna with release-it.
- Add auto-changelog and a changelog handlebars template for formatting the changelog.
- Rename the root CHANGELOG.md to HISTORY.md
- Add a CHANGELOG.md to the cfpb-design-system package, with version v3+.

## Testing

1. In `package.json`, add `--dry-run` to the `release` script and then run `yarn release` and it should allow you to select the version to release.

## Notes

- I'm ignoring the merge title pattern that decap adds, to avoid putting decap edits in the package changelog. I don't love depending on the merge title, but that seemed to be the only way to do it with auto-changelog. 
- I wasn't getting the GitHub release notes to populate correctly, but I wasn't sure if this was because it was a dry run. We'll likely want to do some adjustments after doing some live releases.


contolini (approved):
I'm excited to try this!

PR #2070, opened 2024-10-16
Update Cypress command

anselmbradford:
## Changes

- Update Cypress command to match that in cfgov.

## Testing

1. `yarn cy` should run Cypress.
2. `yarn cy open` should open Cypress.


contolini:
We should prob wait until https://github.com/cfpb/design-system/pull/1916 is merged to verify our `node_modules` dir behaves the same way.

contolini (approved):
🚀

PR #2071, opened 2024-10-17
Replace Avenier Next with Public Sans

contolini:
Just an experiment. 🧑‍🔬


anselmbradford:
Not moving forward Public Sans

Issue #1965, opened 2024-04-25
Icons: Are they always decorative?


contolini:
Are our SVG icons *always* decorative? If so, we should add `aria-hidden="true"` to the SVGs' source code. So far, I haven't been able to find an SVG icon on cf.gov that isn't also coupled with a link, heading, etc.

The other option would be to add a `hidden` parameter to our [svg loading function](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/core/templatetags/svg_icon.py) so that imported SVGs can optionally have `aria-hidden="true"` when desired.

See https://css-tricks.com/accessible-svg-icons/

anselmbradford:
If we do add an attribute to the SVGs, we should be able to add it via svgo by editing the config at https://github.com/cfpb/design-system/blob/main/svgo.config.js and adding something like https://svgo.dev/docs/plugins/add-attributes-to-svg-elements/

sonnakim:
Yes? @willbarton recently generated an SVG icon audit that I can review to check if there are any non-decorative SVGs. 

If SVGs are only decorative, then we can add an aria-hidden attribute using perhaps the script mentioned by @anselmbradford in the comment above

sonnakim:
@anselmbradford So it looks like there are several instances of the warning icon without an accompanying text label on the site. For just three examples: 

https://www.consumerfinance.gov/owning-a-home/conventional-loans/ 
https://www.consumerfinance.gov/about-us/blog/african-american-and-hispanic-borrowers-harmed-provident-will-receive-9-million-compensation/ 
https://www.consumerfinance.gov/about-us/blog/new-relief-for-renters/

I'd argue that these warning icons as used here are not decorative, as they're calling attention to information that the writer believes is important, and there is no accompanying text label to indicate as such, like "Warning" or "Note".

anselmbradford:
@sonnakim It looks like those three cases have vestigial svgs that need to be handled differently post- @willbarton's work. If I view each in the Wagtail admin, the icons disappear, which means they need to be removed anyway, right? Or rather, they'll automatically be removed the next time each of those pages is saved.

That said, if we do keep icons in this way, perhaps we could `aria-hidden` the icons by default, and add hidden text for the cases where it is not decorative. So something like:

```
<div>
   <svg aria-hidden="true" class="cf-icon-svg … >
   <span class="u-visually-hidden">Note:</span>
</div>
```

[...]

PR #2082, opened 2024-10-29
Add `aria-hidden:true` to SVG icons

anselmbradford:
Fixes https://github.com/cfpb/design-system/issues/1965

Adjust SVGo config to add `aria-hidden: true` to SVG icons.

## Changes

- Add `aria-hidden:true` to SVG icons

## Testing

1. Icons should look like before, but when inspected would have an aria-hidden attribute.


contolini (approved):
🪄

PR #2084, opened 2024-10-30
Migrate Lighthouse config from CommonJS to ES6 module

anselmbradford:
## Changes

- Migrate Lighthouse config from CommonJS to ES6 module

## Testing

1. Lighthouse GH action should run as before.


anselmbradford:
Ergh this is unsupported as of yet. See https://github.com/GoogleChrome/lighthouse-ci/issues/973

PR #2085, opened 2024-10-31
Add button link example | Correctly vertically align button group link buttons

anselmbradford:
If a button is accompanied by a button link, the `m-btn-group` sets them to flexbox, but doesn't vertically align them. 

This PR sets the vertical alignment to `align-items: center`. 

## Changes

- Add button link example to buttons page.
- Correctly vertically align button group link buttons.
- Set the button hover color, so that if a link is set as a button, don't change the button text on hover.

## Testing

1. See the button link example at https://deploy-preview-2085--cfpb-design-system.netlify.app/design-system/components/buttons#button-link


packages/cfpb-design-system/src/components/cfpb-buttons/button-group.scss
```diff
@@ -5,4 +5,5 @@
   display: flex;
   gap: math.div(10px, $base-font-size-px) + rem;
   flex-wrap: wrap;
+  align-items: center;
```
anselmbradford, line 8: We could also consider `align-items: baseline` for this, so that the button text and link text bottoms are aligned with each other.

contolini: IMHO `baseline` looks better

contolini (approved):

PR #2095, opened 2024-11-05
Remove `u-break-word`

anselmbradford:
`u-break-word` is only used in the prepaid agreements, which defines its own `u-break-word` with a different behavior! https://github.com/cfpb/consumerfinance.gov/blob/b6f42cebed71a6e9cb58954ef3ce62e21da8cfaf/cfgov/unprocessed/apps/prepaid-agreements/css/main.scss#L13

## Removals

- Remove `u-break-word`

## Testing

1. `u-break-word` should be banished from the codebase and docs!


contolini (approved):
❌

PR #2097, opened 2024-11-05
Add `u-small-text--subtle`

anselmbradford:
This class is used in TCCP.

## Additions

- Add `u-small-text--subtle` modifier for making gray small text.

## Testing

1. There should be mention of the modifier in the docs at https://deploy-preview-2097--cfpb-design-system.netlify.app/design-system/development/helper-classes-and-mixins


contolini (approved):
➕

PR #2103, opened 2024-11-07
Add `a-btn--hide-icon` class

anselmbradford:
When a button with an animating "updating" icon is used, the icon would generally be hidden when the action is complete. However, the DS did not have a standard way of hiding the icon and applicables set their own CSS to accomplish this (e.g. https://www.consumerfinance.gov/privacy/records-access/). This PR adds a class that hides the icon (and the icon divider!)

## Additions

- Add `a-btn--hide-icon` class

## Testing

1. Visit the buttons page in the PR preview and edit any button with icon CSS to add `a-btn a-btn--hide-icon` and see that the icon and divider magically disappear! 🪄


contolini (approved):
Normally I'd advocate for just using the `u-hidden` utility class but it's the nested SVGs that need targeting so this looks good 👍

PR #2118, opened 2024-11-15
Add `u-w65pct` and `u-w15pct` utilities

anselmbradford:
`u-w65pct` and `u-w15pct` are used on the activity-log page. Probably we should make a dedicated table component for that and get rid of these. But in the meantime, this consolidates the width utilities together.

## Changes

- Add `u-w65pct` and `u-w15pct` utilities

## Testing

1.

## Screenshot
Screenshot 2024-11-15 at 6 11 29 PM


contolini (approved):
One step closer to tailwind 👍

PR #2121, opened 2024-11-20
Tweak and consolidate date picker

anselmbradford:
Merge two date picker examples we had, and make a regular and full width example.

## Changes

- Tweak and consolidate date picker examples

## Testing

1. Check out the examples on across major browsers and iOS simulators https://deploy-preview-2121--cfpb-design-system.netlify.app/design-system/components/date-picker


packages/cfpb-design-system/src/base/base.scss
```diff
@@ -35,6 +35,10 @@ textarea {
 input[type='date'] {
   appearance: textfield;
   display: revert;
+
+  &.a-text-input--full {
+    -webkit-min-logical-width: calc(100% - 16px);
```
contolini, line 40: Why not `min-width`?

anselmbradford: Ok ok I went a min-width approach in https://github.com/cfpb/design-system/pull/2121/commits/c41101f6be4994c8532b884191525df269529a9d

I realized the calc(100% - 16px) works on iOS, but then it is shorter on desktop browsers. Instead I set a min width/height to what it would be if the appearance was a textfield, and then set the appearance to none. This appears to look good across iOS Safari, desktop Safari, FF, and Chrome.

contolini (approved):

PR #2130, opened 2024-11-29
Update to yarn 4.5.3 | stylelint 16.11.0

anselmbradford:
According to https://yarnpkg.com/configuration/yarnrc#yarnPath, `yarnPath` isn't the recommended way to install yarn in a project any longer, and it should instead be handled by CorePack.

## Changes

- Update to yarn 4.5.3 
- stylelint 16.11.0

## Testing

1. PR checks should pass.


contolini (approved):
Makes sense--it's how we install yarn locally on our machines

PR #2139, opened 2025-01-07
Provide instantiated tooltip object on initialized component

wpears:
Some consumers of these tooltips may need access to the underlying tippy object to, eg, check the tooltip state, etc.
This PR provides the initialized tippy object as a `tooltip` key on the initialized component instance (this key is null before initialization).

### DIFF
--- a/packages/cfpb-design-system/src/components/cfpb-tooltips/index.js
+++ b/packages/cfpb-design-system/src/components/cfpb-tooltips/index.js
@@ -22,7 +22,7 @@ function Tooltip(element) {
 
   /**
    * Set up and create the tooltip.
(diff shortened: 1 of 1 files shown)


packages/cfpb-design-system/src/components/cfpb-tooltips/index.js
```diff
@@ -25,7 +25,7 @@ function Tooltip(element) {
    * @returns {Tooltip} An instance.
```
anselmbradford, line 25: This doc isn't quite right, is it? The returned type is whatever `tippy` returns, not an instance of `Tooltip`?

PR #2141, opened 2025-01-10
Update release script for more reporting and checking

anselmbradford:
## Changes

- Update release script for more reporting and checking.

## Testing

1. Unset `GITHUB_TOKEN` and run `./scripts/release.sh` and it should now prompt you if `GITHUB_TOKEN` is not set.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,9 +1,35 @@
 #!/usr/bin/env bash
 
-yarn lint &&
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m'
 
-yarn build &&
+if [ ! -z "${GITHUB_TOKEN}" ]; then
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -1,9 +1,35 @@
 #!/usr/bin/env bash
 
-yarn lint &&
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m'
 
-yarn build &&
+if [ ! -z "${GITHUB_TOKEN}" ]; then
```
wpears, line 7:
```suggestion
if [ -z "${GITHUB_TOKEN}" ]; then
```

Double negative!

wpears (changes requested):

scripts/release.sh
anselmbradford: Fixed in https://github.com/cfpb/design-system/pull/2141/commits/5ff5dd63b039e19faf42057d97daabd5e74f85e3

wpears (approved):
🆒

PR #2186, opened 2025-04-11
Notifications: Set max-width of content

anselmbradford:
Notification content should wrap if it gets too wide.

## Changes

- Notifications: Set max-width of content.

## Testing

1. Check PR preview for the alerts page and edit the notification text to be really long and see that it wraps before the width of the notification background/border.

Screenshot 2025-04-11 at 6 28 01 PM


natalia-fitzgerald (approved):
@anselmbradford 
This looks great!

PR #2192, opened 2025-04-18
Buttons: Adjust secondary button hover color

anselmbradford:
## Changes

- Buttons: Adjust secondary button hover color

## Testing

1. Check the secondary button styles and see that the hover color is pacific-10.


natalia-fitzgerald (approved):
@anselmbradford - This looks great! Approved.

Issue #2207, opened 2025-05-09
Bug: Filter tag "active" state color


natalia-fitzgerald:
**Describe the bug**
There is an error in the active state background color for the filer tag. The active state background color should be Teal 60 and it is currently coded as Teal. The combination of Teal and Black do not meet contrast ratio requirements.

Here's the original PR where this was added to the CFPB Design System with the active color background set to Teal 60. https://github.com/cfpb/design-system/pull/1777

Issue for context: https://github.com/cfpb/design-system/issues/1760

**To Reproduce**
1. Go to '[filter tag](https://cfpb.github.io/design-system/components/tags#filter-tag)'
2. Click on one of the filter tags and hold down
4. See background color and observe the poor contrast ratio between background and text

**Expected behavior**
Background color should be Teal 60 while in the active state.

**Screenshots**

Bug
Active state background color is set to Teal
<img width="238" alt="Image" src="https://github.com/user-attachments/assets/ffeb7d84-ea9c-40c5-8243-7b5c1cb72204" />

Correct
Active state background color should be set to Teal 60
<img width="240" alt="Image" src="https://github.com/user-attachments/assets/71e6717b-8185-4c4a-8830-b6bb1fb8263e" />

**Additional context**

@sonnakim for awareness

PR #2220, opened 2025-05-09
Update the a-tag-filter active color to teal-60 from teal

anselmbradford:
Fixes https://github.com/cfpb/design-system/issues/2207

## Changes

- Update the a-tag-filter active color to teal-60 from teal

## Testing

1. Check tags in the PR preview and see that the active color is correct.
https://deploy-preview-2220--cfpb-design-system.netlify.app/design-system/components/tags#filter-tag


natalia-fitzgerald (approved):
@anselmbradford - Looks great!

PR #2237, opened 2025-05-14
Remove `m-list--spaced`

anselmbradford:
It doesn't appear we use this modifier.

## Removals

- Remove `m-list--spaced`

## Testing

1. PR checks should pass and spaced list example should be gone from preview site.


itsmedavep:
Page looks good and removal makes sense!~

itsmedavep (approved):
This all looks good to me! Thank you for the update/removal.

PR #2264, opened 2025-05-30
Add `lit` and initial web components

anselmbradford:
## Additions

- Add a new "elements" page with initial web components: cfpb-button and cfpb-file-upload.
- Add `lit` dependency for web components.

## Testing

1. Check components > elements page for example web components.

## Screenshots

Screenshot 2025-05-30 at 7 27 23 PM


anselmbradford:
Regular components /dist/ files are receiving uncompressed dupes of the web components code. Lemme know if you see why.

wpears (commented):
This looks directionally correct to me! I do wonder about trying to isolate scss from component code. That is, instead of having component scss with `use` statements that gets loaded as a text, what if we put the translation from scss-land into lit css template tags higher up, that is, in `core`, `vars`, etc. Then in component code we could simply import { vars } from './where/ever.js' and use them as normal in lit.

There's kinda a delicate dance, I'll admit, between maintaining scss for child consumers (mostly cf.gov) and making the lit components more idiomatic. Not sure what the best thing is here.

As to the practicalities of this PR, looks like there's some commented stuff in the file-upload index.js file we could trim.

anselmbradford:
> Regular components /dist/ files are receiving uncompressed dupes of the web components code. Lemme know if you see why.

I figured this out. Internally within the DS packages, we had the convenience import of, e.g., `import { BaseTransition, EventObserver } from '@cfpb/cfpb-design-system';`. This imports from the main DS index file, which exports all public JS methods, including the web component code. This led to the web component code ending up in the other non-web component JS bundles. If we instead update the convenience imports to the actual locations of the JS methods, then the individual component bundles are scoped only to their immediate dependencies. See https://github.com/cfpb/design-system/pull/2264/commits/09f043437ec290990d690571fa971dd6ddfc7223

anselmbradford:
I'm going to merge this and open a separate PR that tweaks the optimizations. `cssnano` appears to provide slightly better compression than `postcss-minify`, and also might be able to handle autoprefixer as well. It would also be nice to remove whitespace from the web component markup in the JS file. It doesn't matter too much right now, because the markup is minimal, but could make a bigger difference in more complex components in the future.

PR #2265, opened 2025-06-02
Fix Lighthouse failures

chosak:
Nightly Lighthouse tests against the design system website are currently failing due to issue with two pages:

https://github.com/cfpb/design-system/actions/runs/15380846769/job/43271344392

This PR fixes those Lighthouse failures.

Note that two of the links being corrected here no longer work.


anselmbradford:
> Note that two of the links being corrected here no longer work.

Should they be updated to wayback archive URLs? I'll merge this to fix the lighthouse failures, but the URLs should be edited in decap.

PR #2285, opened 2025-06-05
Remove legacy form input CSS

anselmbradford:
## Changes

- Remove legacy IE autoprefixer ignore line to inline-table.
- Remove `a-label__helper`, which no longer appears to change anything.

## Testing

1. Check the reference page and see that the labels with form inputs are on their own line.
https://deploy-preview-2285--cfpb-design-system.netlify.app/design-system/development/reference-for-component-states


wpears:
Yeah seems like `.a-label__helper--block` is what this behavior is trying to match

wpears (approved):

Issue #2234, opened 2025-05-14
Filter tag using an anchor doesn't have a hover state


anselmbradford:
**Describe the bug**

**To Reproduce**
1. Go to https://cfpb.github.io/design-system/components/tags#filter-tag
2. Hover over last tag example
3. See that there is no color change

**Expected behavior**
There should be a color change on hover

@sonnakim for awareness

PR #2296, opened 2025-06-12
Move tags to reference page

anselmbradford:
## Changes

- Move tags to reference page.
- Fix erroneous markup on the tag guidance page.
- Fix hover state on filter tag that is on an anchor link.

## Testing

1. Check the tags on the PR preview reference page.


anselmbradford:
Fixes https://github.com/cfpb/design-system/issues/2234

PR #2301, opened 2025-06-13
Optimize web component code

anselmbradford:
## Changes

- Replace `postcss-minify` with `cssnano`
- Add `esbuild-plugin-minify-html-literals` for compressing HTML literals in the bundle.
- Move `autoprefixer` to `pluginPostCssSass` to remove redundant import.

## Testing

1. PR preview legacy components (e.g. expandables) should still work. Web components should work as before. 
2. Built /dist/ .js files should have compressed html literals.


anselmbradford:
`cssnano` has an "advanced" preset that includes autoprefixer (see https://cssnano.github.io/cssnano/docs/what-are-optimisations/#what-optimisations-do-you-support%3F), however, it requires installing `cssnano-preset-advanced`, so we wouldn't save a dependency by using it, and it turns on some other potentially breaking config, so it's probably better to use autoprefixer directly for now.

esbuild/docs/scripts.js
```diff
@@ -20,7 +19,7 @@ function scripts(baseConfig) {
     target: 'es6',
     plugins: baseConfig.plugins.concat([
       pluginPostCssSass({
```
wpears, line 21: Can this just become `pluginPostCssSass()`?

PR #2303, opened 2025-06-13
Add static init methods to WC

anselmbradford:
Currently, we import our components and then call an `.init()` method on them to initialize them. The web components initialized without this method. To fit into the existing paradigm, this PR adds static init methods to the WC definition files. 

## Changes

- Add static init methods to WC.

## Testing

1.  https://deploy-preview-2303--cfpb-design-system.netlify.app/design-system/components/elements should be unchanged.


anselmbradford:
Works for me! We may want to optimize where these are initialized in the future. At very least, we should probably keep the check on whether it has already been registered, otherwise we get a console error. 


Closing this in favor of further polish in https://github.com/cfpb/design-system/pull/2304

PR #2309, opened 2025-07-01
Replace Avenir Next with Source Sans 3

anselmbradford:
## Changes

- Swap Avenir Next with Source Sans 3 Variable.
 
## Testing

1. See PR preview and check the pages to see that nothing looks broken https://deploy-preview-2309--cfpb-design-system.netlify.app/design-system/foundation/fonts


anselmbradford:
> I've been kinda perturbed at /dist/ getting built and included in main altogether. Is this for netlify previewing or packaging?

IIRC it's to allow accessing the built files from the package without needing to necessarily run them through a build step. For example, here https://github.com/cfpb/ccdb5-ui/blob/main/src/css/base.scss#L4 — although that's not the best example, as that still has a build step involved. However, it does use the /dist/ file.

The aim was to have these in the published npm package, so if there's a step in the publish to npm process that builds the /dist/ files before publishing, that may be preferable.

wpears:
> so if there's a step in the publish to npm process that builds the /dist/ files before publishing, that may be preferable.

That's exactly what I was thinking, noted in #2314

jenn-franklin (approved):

PR #2337, opened 2025-08-12
Add tag web components

anselmbradford:
This PR adds Web Component versions of the topic and filter tags. This is a precursor to a web component multiselect component. 

## Additions

- Add tag web components

## Testing

1. Check the topic and filter tags on https://deploy-preview-2337--cfpb-design-system.netlify.app/design-system/components/elements and see that they work on hover and tabbing (note that first topic tag is not a link).


itsmedavep:
I see a differences here with the non web component version of the topic tag. Looks to me like font-size is:14px for the non component.

Non web component version:
Screenshot 2025-08-27 at 7 36 23 AM

Web component version:
Screenshot 2025-08-27 at 7 35 43 AM

itsmedavep:
Filter tags:

Font size differences. 

Non web component:
Screenshot 2025-08-27 at 7 44 35 AM

Web component:
Screenshot 2025-08-27 at 7 44 17 AM


SVG close icon differences:

Non web component:
Screenshot 2025-08-27 at 7 49 31 AM

Web component:
Screenshot 2025-08-27 at 7 49 10 AM


Row gap:

IDK if this matters but the non web component version uses REM and the web component version uses px for the row gap (distance between text and icon).

anselmbradford:
I added a `addTag` and `removeTag` API to the tag group.

itsmedavep:
Just noticed something playing with the example. 

Seems like if I add topic tag (no link) but then click on it it removes the tag. That also shows in the log view. IDK if that is intended behavior or not but thought that I should point it out. 

A user is not likely to click on it since its not a link but then we probably don't want it removed from the page.  Otherwise this is really cool!

PR #2389, opened 2025-10-03
Move link underlines to text-decoration instead of borders

anselmbradford:
## Changes

- Move link underlines to text-decoration instead of borders.

## Testing

1. Pull branch and `yarn start` and visit http://localhost:4000/design-system/development/reference-for-component-states and check the links across screen sizes.


itsmedavep (approved):
Verified that these changes all look good in the DS site. Verfied that they are all using the text-decoration properties now instead of border.

Issue #2253, opened 2025-05-22
Decap Page Title editor not allowing spaces


anselmbradford:
**Describe the bug**
When editing a page, it's not possible to write in the page title, it's not possible to write spaces. This is probably because of a bug in `StringTrimmed.js`.

**To Reproduce**
1. Go to a page in decap CMS
2. Edit the page.
3. Try to edit the page title with multiple words and see that the spacebar doesn't work.

**Expected behavior**
Should be able to add multiple word page titles.

@sonnakim for awareness

PR #2406, opened 2025-10-17
Fix trim on inputs

anselmbradford:
Fixes https://github.com/cfpb/design-system/issues/2253

## Changes

- Makes trim happen on blur and properly sets it, as `trim()` isn't mutable on the source value.

## Testing

1. Pull branch and `yarn start` and edit a page in decap and see that you can add spaces to the page title and when you click off the input the extra spaces are trimmed off.


itsmedavep (approved):
Pulled the branch and built. Verified that adding spaces to a title and then clicking outside of the filed in CMS trimmed spaces. 

All looks good to me.

PR #2527, opened 2026-02-25
Add Cypress component testing for lit

anselmbradford:
Draft branch with work needed to run lit components in Cypress Component Tests. However, we may not want to proceed with this if we can do testing inside Storybook. 

Note: Vite config needs to go in first for this to work.

## Additions

- Add `cypress-ct-lit` to mount lit components in Cypress.

## Changes

- Move Cypress directory to /cypress/

## Testing

1. `yarn cy open` and run component test.


anselmbradford:
Closing this out, as we've replaced Cypress with Playwright, and have a Storybook installation running now.

PR #2574, opened 2026-03-25
Consolidate elements JS imports

anselmbradford:
## Removals

- Remove `/elements/base/index.js` that was importing only scss

## Changes

- Consolidate elements JS imports into `/elements/index.js`

## Testing

1. `yarn build && yarn start` should pass and the doc site should be unchanged.


itsmedavep (approved):
I pulled this and built. No visual regressions as far as I can tell. 

Approved!

PR #2597, opened 2026-04-03
Tweak test to type one character instead of four

anselmbradford:
This test is very brittle and often fails with only 'bes' being typed instead of 'best'. This adjusts the test to type one emoji, which should be less likely to fail the test before it has finished typing, and is also unlikely to be a false positive against the page's existing content. 

## Changes

- Tweak test to type one character instead of four
 
## Testing

1. PR checks should pass.


itsmedavep (approved):
This looks good. I pulled the branch started up the local server and then ran `yarn cy` All tests passed.

PR #2598, opened 2026-04-03
cfpb-search-form: Hide autocomplete list on page load

anselmbradford:
The autocomplete list for the cfpb-search-form was flashing on page load, and also appearing when JS was disabled. This adds a `hidden` attribute to the slotted autocomplete list.

## Changes

- cfpb-search-form: Adds a `hidden` attribute to the slotted autocomplete list.
- Add FOUC section to fallback docs.

## Testing

1. `yarn build` and visit http://localhost:4000/design-system/components/reference-for-custom-elements and see the search form doesn't flash a drop-down list on page load.


itsmedavep (approved):
I pulled this branch and built. I no longer see the autocomplete list on page load or when JS is disabled.

Approved!

PR #2611, opened 2026-04-08
Add loading skeleton architecture

anselmbradford:
A "loading skeleton" is a placeholder box before a web component has loaded in.

## Changes

- Add loading skeleton architecture.
- Fix issue where tagline was loosing top/bottom padding at mobile screen size.

## Testing

1. Find `CfpbTagline.init();` and `CfpbFormSearch.init();` in `index.js` and comment them out.
2. `yarn build && yarn start`
3. Visit http://localhost:4000/design-system/ 
4. See that there are boxes where each component should be. 
5. Undo step 1, and re-run step 2-3.
6. See that the components load in normally. See that there isn't a layout shift when they load in.

Screenshot 2026-04-08 at 4 42 24 PM


packages/cfpb-design-system/src/elements/cfpb-form-search/skeleton.css
```diff
@@ -0,0 +1,8 @@
+cfpb-form-search:not(:defined) {
+  min-height: 35px;
+
+  /* Mobile only */
+  @media only screen and (width <= 37.5625em) {
```
anselmbradford, line 5: I didn't use the SASS mixin here because I wanted to keep this in `.css` with the idea we'll aim to nix the media-query mixins anyway.

packages/cfpb-design-system/src/elements/cfpb-utilities/skeleton.css
```diff
@@ -0,0 +1,33 @@
...
+@keyframes pulse-background {
+  0% {
+    background-color: transparent;
+  }
+
+  50% {
+    background-color: rgb(90, 93, 97, 5%);
+  }
+
+  100% {
```
itsmedavep, line 30: Just a FYI,  I don't think that you need to include the 100% keyframe if it has the same value as the 0% one. Though it may be good for covering older browsers.

itsmedavep (approved):
I was able to pull the branch and test out this approach. Looks good. Approved!

packages/cfpb-design-system/src/elements/cfpb-utilities/skeleton.css
```diff
@@ -0,0 +1,33 @@
...
+
+  box-shadow: 0 0 3px rgb(90, 93, 97, 10%) inset;
+  border-radius: 4px;
+
+  animation: pulse-background 1.5s infinite;
+}
+
+@keyframes pulse-background {
+  0% {
+    background-color: transparent;
```
itsmedavep, line 23: I guess one further thought..... I might consider not having this go all the way transparent at the 0 point. Its very subtly there and might be easy to miss. 

IDK how much that matters TBH.

anselmbradford: I'll merge this in to get the architecture in, and we can iterate on the actual skeleton appearance.

PR #2613, opened 2026-04-09
Use resolutions field to update `trim`

anselmbradford:
`trim` has a security vulnerability down our dependency tree. This bumps the version in the resolutions field in our package.json.

 `trim` is a `decap-cms-app` transitive dependency.

## Changes

- Use resolutions field to update `trim`

## Testing

1. PR checks should pass. Keep an eye out for issues editing in decap.

## Notes

I tried bumping `qs` too, which is a cypress, decap-cms-app, style-dictionary, lhci/cli (lighthouse), and http-server (local run server for lighthouse) transitive dependency. However, doing that made it so lighthouse produces a Chrome error and no longer passes.


anselmbradford:
Moved to https://github.com/cfpb/design-system/pull/2614

PR #2623, opened 2026-04-10
Add `cfpb-icon`

anselmbradford:
Adds a `cfpb-icon` that lazy loads icons via a dynamic `img` tag with a mask layered on top (to set the color).

1. In whichever implementation, the path to the icons are passed to  `CfpbIcon.init('/path/to/icons/')`.
2. `img` added in the web component shadow DOM with a source of the whichever icon is set in the `name` attribute.
3. Wrap the `img` in a `span`.
4. The span sets `mask-image` with a source of the same icon path, in order to load in a mask of the same icon on top of the icon image.
5. The `background-color` is set to a custom variable that uses whatever color name is passed through the `color` attribute (within the CFPB color scheme).

Upsides:
- No need to deal with strict mime type settings on the server.
- Lazy load icons using in-browser image/CSS optimizations.

Downsides:
- Two requests per icon (though I think this should cache?).
- Loss of access to underlying SVG (no `fill` able to target the SVG).

## Additions

- Add `cfpb-icon` web component.

## Testing

1. `yarn build && yarn start` and visit http://localhost:4000/design-system/components/reference-for-custom-elements and see the icons at the top. Inspect them and change the `color` and `name` attributes to change the icon name and color. 

## Screenshots

Screenshot 2026-04-10 at 6 19 25 PM


itsmedavep (approved):
I pulled this branch and built. I verified the behavior. Looks good to me! 

Approved.

PR #2633, opened 2026-04-16
Migrate cfpb-checkbox-icon to use cfpb-icon

anselmbradford:
## Changes

- Migrate cfpb-checkbox-icon to use cfpb-icon
- Add `em-from-px` shared function.

## Testing

1. `yarn build && yarn start` and visit the elements reference page and check that the checkboxes (checkbox inputs and multiselect drop-down) are aligned correctly (with and without focus).

## Screenshots

| Before | After  |
| ------ | ------ |
|Screenshot 2026-04-16 at 10 55 51 AM|Screenshot 2026-04-16 at 10 55 40 AM|

## Notes and todos

- This update collapses the space around the checkboxes. This means if placed side-by-side, the outlines around the checkboxes overlap. This is correct behavior in terms of how outlines work, but may not be the design intent, if the outlines are actual faux borders.


itsmedavep (approved):
I pushed an update to the abstracts/vars.json and then generated elements/abstracts/vars.css. 

That overwrites your manual changes to that file in favor of the SD generated file from the json. 

Then I built the branch and verified that everything still works. 

Image

It did and everything still looked good. 

Approved!

PR #2693, opened 2026-06-03
Remove legacy `a-text-input__full`

anselmbradford:
This class appears to be unused.

## Removals

- Remove legacy `a-text-input__full`

## Testing

1. `yarn build && yarn start` and visit http://localhost:4000/design-system/ and the site search input should still be full-width at mobile.


flacoman91 (approved):
Thank you!

PR #2751, opened 2026-07-29
Temporary fix for issue where Firefox is rendering font heights differently from Chrome

anselmbradford:
We need to set `font: inherit` for form controls (button, input, select, textarea) to override the user agent stylesheet. However, in inputs this appears to actually break things. Firefox renders the incorrect height for the input and the wrong font. Removing this line fixes it and makes Firefox and Chrome consistent.

## Changes

- Remove `font: inherit` from search input.

## Testing

1. `yarn build && yarn start` and the search input in firefox is not too tall.

## Screenshots

| Before | After  |
| ------ | ------ |
|Screenshot 2026-07-29 at 8 27 18 AM|Screenshot 2026-07-29 at 8 27 21 AM|


anselmbradford:
The issue is actually that the button isn't Source Sans 3, but the fallback arial instead, since `cfpb-form-search` also needs to set `font: inherit` for the button. This PR will fix the height issue, but the search input and button text will both render in the fallback arial typeface. I'll merge this to fix the size issue, but it needs further attention to get both the input and button to render source sans 3.

PR #2781, opened 2026-08-08
Fix search input height and font-family

anselmbradford:
In Firefox and Safari, the height of the search input is too high in comparison to the height in Chrome. This PR sets the search input and button to Source Sans 3 Variable (it's currently erroneously Arial), and sets the input container to a flexbox layout in combination with a relative height, which fixes the height. It also sets the input values to relative units, so that it can properly resize if the user were to change their base font.

## Changes

- Fix search input height and font-family.
- Convert fixed units to relative units to fix issues when user changes base font size.

## Testing

1. `yarn build`
2. Check http://localhost:4000/design-system/components/reference-for-custom-elements in both Firefox, Safari, and Chrome. The height of the search input at the top of the page should be the same height as the associated button. 
3. Check http://localhost:4000/design-system/components/text-inputs and see that the legacy search input has the same input height in Firefox, Safari, and Chrome.
4. You can go into your browser preferences and change the base font size and see that the search input handles the changes better than before.

## Notes and todos

- If the base font is set smaller than 16px, the component gets kinda messed up, particularly in regard to the autocomplete drop-down. I imagine we don't care about this too much and should prioritize if the base font size is set larger than 16px.

- We could set the min-width of the input (currently 300px) to a rem unit, so that the input doesn't change width when interacted with when the base font size is larger than 16px. However, this'll bust the input layout when the base font size is under 16px, so I left this as a fixed 300px for now.


wpears (approved):
🪄

PR #2786, opened 2026-08-13
Update e2e so that we cache Playwright browser

itsmedavep:
Cache browsers to help speed up CI runs. This should provide us with the structure to allow for caching Playwright browsers going into the future should our list expand. 

What you should see is that the first push on each branch is a cache miss and headless browser downloads shell only. Every run after on the same Playwright version restores from cache and skips the install step. The cache key will change when we bump `@playwright/test`. 

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Screenshots

| Before | After  |
| ------ | ------ |
|        |        |

## Notes and todos

-

## Checklist

### Browsers

Check the [current browser support cutoff list](https://cfpb.github.io/consumerfinance.gov/browser-support#current-browser-support-metrics) for browsers that are advisable
to prioritize for testing.

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -15,8 +15,25 @@ jobs:
       - name: Shared setup
         uses:  ./.github/actions/shared-action-setup
 
-      - name: Install Playwright browsers
-        run: npx playwright install --with-deps chromium
+      - name: Resolve Playwright version
+        id: playwright-version
+        run: echo "version=$(node -p "require('@playwright/test/package.json').version")" >> "$GITHUB_OUTPUT"
+
+      # Browser binaries live outside node_modules. Key on the exact
+      # Playwright version. A version bump must miss the cache, so don't restore
+      - name: Cache Playwright browsers
+        id: playwright-cache
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0 https://github.com/actions/cache/releases
+        with:
+          path: ~/.cache/ms-playwright
+          key: ${{ runner.os }}-playwright-${{ steps.playwright-version.outputs.version }}-chromium-shell
+
+      # --only-shell downloads just the chromium headless shell
+      # which is what all these headless tests need. If a run ever
+      # fails on a missing shared lib, add a step `npx playwright install-deps chromium`
+      - name: Install Playwright chromium
+        if: steps.playwright-cache.outputs.cache-hit != 'true'
+        run: npx playwright install --only-shell chromium 
 
       - name: Run Playwright tests
-        run: npx playwright test
+        run: npx playwright test --project=chromium


.github/workflows/e2e.yml
```diff
@@ -15,8 +15,25 @@ jobs:
...
         uses:  ./.github/actions/shared-action-setup
 
-      - name: Install Playwright browsers
-        run: npx playwright install --with-deps chromium
+      - name: Resolve Playwright version
+        id: playwright-version
+        run: echo "version=$(node -p "require('@playwright/test/package.json').version")" >> "$GITHUB_OUTPUT"
+
+      # Browser binaries live outside node_modules. Key on the exact
+      # Playwright version. A version bump mus miss the cache, so don't restore
```
anselmbradford: "mus" typo?

itsmedavep: Whooops. Nice catch. TY!

.github/workflows/e2e.yml
```diff
@@ -15,8 +15,25 @@ jobs:
...
+      - name: Cache Playwright browsers
+        id: playwright-cache
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # version 6.1.0 https://github.com/actions/cache/releases
```
anselmbradford: Let's format the comment as `# v6.1.0` (appended URL is fine).
This is the format in other pinned action comments, so that way you can search the codebase for ` # v` and find them all.

itsmedavep: Sounds good! TY.

.github/workflows/e2e.yml
```diff
@@ -15,8 +15,25 @@ jobs:
...
+      # --only-shell downloads just the chromium headless shell
+      # which is what all these headless tests need. If a run ever
+      # fails on a missing shared lib, add a step `npx playwright install-deps chromium`
+      - name: Install Playwright chromium
+        if: steps.playwright-cache.outputs.cache-hit != 'true'
+        run: npx playwright install --only-shell chromium 
 
       - name: Run Playwright tests
-        run: npx playwright test
+        run: npx playwright test --project=chromium
```
anselmbradford, line 39: What would be the changes if we wanted to add Firefox checks, for example? Easy enough to add?

itsmedavep: @anselmbradford Its a bit of work but doable. I think in that case 4 things would need to change:

1.  `playwright.config.ts` - Uncomment the Firefox stuff: [line: 48](https://github.com/cfpb/design-system/blob/d3a381f697526907976e85de2603937e6767e330/playwright.config.ts#L48) which should be the only change needed outside the action
2. The cache key must change. If we added Firefox to the install command but left the key as `-chromium-shell` the existing cache would restore and the cache-hit would be `true` so Firefox would never d/l. Funky potential bug ... if that wound up being the case then the tests would fail with a browser launch that looked correct. 
3. `--only-shell` only applies to Chromium.... Firefox has no headless shell as far as I can tell (ugh) so we would get the full build. So Firefox would cost us some time and MB on every cache miss.
4. System deps come back. Right now this way works and should be fast because Chromium's libs ship on the runner. Playwright's Firefox build needs packages that aren't all there. 

When/if we do the update to include Firefox

It sort of becomes a matrix so chromium doesn't wait on Firefox.

```shell
strategy:
     fail-fast: false
     matrix:
          include:
               - project: chromium
                 install_flags: '--only-shell'
               - project: firefox
                 install_flags: ''
```

Then the cache key would become something like: `${{ runner.os }}-playwright-${{ steps.playwright-version.outputs.version }}-${{ matrix.project }}` and then the install step would run `npx playwright install ${{ matrix.install_flags }} ${{ matrix.project }} ` Then a new step would be added for Firefox which would add the deps it needs.  Then the test step would run `--project${{ matrix.project }}`

I have not verified that Firefox doesn't wind up with all that it needs on the runner. If we found that to be the case, then we would not need the step that installs them.

itsmedavep: We opted to fold this in for now and ignore the Firefox considerations for the time being.

anselmbradford (approved):

PR #2794, opened 2026-08-25
Fix hover for cards that span two lines

anselmbradford:
If a card spans two lines, the hover on adjacent cards will have a gap. This fixes that by setting the height to 100% on the inner dimensions of the card.

## Changes

- Fix hover for cards that span two lines

## Testing

1. `yarn build` and add some long text to the breakout cards and see that adjacent cards are the same height.

## Screenshots

Before
Screenshot 2026-08-25 at 6 32 35 PM

After:
Screenshot 2026-08-25 at 6 42 35 PM


niqjohnson (approved):
Confirmed this fixes the cards on the cf.gov homepage when one wraps to two lines and the other doesn't.

itsmedavep (approved):
This looks good to me! Thank you for the review @niqjohnson

PR #2801, opened 2026-09-03
Add Storybook links to the docs site for Web Components and RDS Storybook

itsmedavep:
This addresses the desire to have the docs site link to the WC and Reac Storybooks. 

This only covers the docs pages where we have Storybook stories for both the WC and React versions as a first step in the process. 

Additionally, this adds the Storybook svg icon and colors it with a class by wrapping it in a `<span>`. 

That was necessary because DeCap descriptions use >- style markdown and if we used the cfpb-icon colorization natively it would hit a YAML parsing error. (it works fine in the WC reference docs page because it lives outside of a description)

That means it covers:

alerts
buttons
expandables
filter tags
taglines

`yarn start` and verify there are now links to WC and React Storybooks.

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Screenshots

| Before | After  |
| ------ | ------ |
|        |        |

## Notes and todos

-

## Checklist

### Browsers

Check the [current browser support cutoff list](https://cfpb.github.io/consumerfinance.gov/browser-support#current-browser-support-metrics) for browsers that are advisable
to prioritize for testing.

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


itsmedavep:
| iconography page | alerts page |
|-------------------|------------|
| Screenshot 2026-09-03 at 3 09 41 PM|Screenshot 2026-09-03 at 3 10 04 PM|
| buttons page | expandables page |
|Screenshot 2026-09-03 at 3 10 15 PM|Screenshot 2026-09-03 at 3 10 32 PM|
| filter tags | taglines page |
|Screenshot 2026-09-03 at 3 11 18 PM|Screenshot 2026-09-03 at 3 11 05 PM|

docs/pages/alerts.md
```diff
@@ -8,6 +8,19 @@ description: Alerts draw a user's attention to a change in the status of a form
   the form title. Field-level alerts appear inline with input fields and can
   highlight successful submissions, errors that need to be corrected, or details
   to know before submitting a form.
+
+  <div>
+  <span class="storybook-brand-color"><cfpb-icon name="storybook" /></span>
+  <cfpb-link link-variant="external"><a href="https://cfpb.github.io/design-system/web-components/?path=/docs/web-components-cfpb-alert--overview">Web component</a></cfpb-link>
```
niqjohnson: Can this be marked up as our standard [link with icon component](https://cfpb.github.io/design-system/components/links#link-with-icon)? If we want to go that route, I think we can get rid of the external link icon at the end and use just the Storybook icon as the icon (since these links all still point to cfpb.github.io). I think we can get around having to wrap the icon in a `span` if we move the markup to a template instead of in the markdown.

docs/pages/alerts.md
```diff
@@ -8,6 +8,19 @@ description: Alerts draw a user's attention to a change in the status of a form
   the form title. Field-level alerts appear inline with input fields and can
   highlight successful submissions, errors that need to be corrected, or details
   to know before submitting a form.
+
+  <div>
```
niqjohnson: Can this be marked up as our standard [link list component](https://cfpb.github.io/design-system/components/links#list-link)? Semantically I think that makes sense, but I'm not sure if it'll throw anything off in the layout.

niqjohnson (commented):
High level, I wonder if we can refactor this a little to centralize all the markup for the links in one spot so we don't have to repeat it on every page. Take this with a grain of salt since I'm out of practice with the Design System, though. Can we move the markup to [`variation-content.html`](https://github.com/cfpb/design-system/blob/main/docs/_includes/variation-content.html#L64), something like:

```jinja
{% if page.storybook_links and page.storybook_links.size > 0 %}
    <ul class="m-list m-list--links">
        {% for storybook_link in page.storybook_links %}
            <li class="m-list__item">
                <a class="a-link a-link--jump" href="{{ storybook_link.href }}">
                    [storybook icon goes here]
                    <span class="a-link__text">{{ storybook_link.text }}</span>
                </a>
            </li>
        {% endfor %}
    </ul>
{% endif %}
```

Then we'd have a `storybook_links` field in the markdown files that can contain any number of links in it (the content of which would be the link text and the link URL). That would probably have to be hooked up to appear in Decap, too, right?

So it would be more upfront setup this way, but it would probably make it easier to add more links in the future (and would keep the markup in one spot if we ever need to change it).

What do you think?

docs/pages/alerts.md
itsmedavep: A couple of things

1) I tired to match the concept that Jenn had closely, if we want to move away from that I think what you suggest here makes sense. 
2) I was trying to use web components for this since we really should be moving off the legacy components IMO.
3) Given that, if we loose both icons in favor of only the Storybook one it probably would become `<cfpb-link href="#" icon-left="Storybook">This is a link</cfpb-link>`

docs/pages/alerts.md
itsmedavep: It can and probably should but I believe it introduced layout issues when I tried to use the WC cfpb-list component. I can revisit. 

https://cfpb.github.io/design-system/components/reference-for-custom-elements#list

EG:

```html
<cfpb-list>
  <cfpb-link><a href="#">Item 1</a></cfpb-link>
  <cfpb-link><a href="#">Item 2</a></cfpb-link>
  <cfpb-link><a href="#">Item 3</a></cfpb-link>
</cfpb-list> <cfpb-list gap="compact">
```

itsmedavep:
> High level, I wonder if we can refactor this a little to centralize all the markup for the links in one spot so we don't have to repeat it on every page. Take this with a grain of salt since I'm out of practice with the Design System, though. Can we move the markup to [`variation-content.html`](https://github.com/cfpb/design-system/blob/main/docs/_includes/variation-content.html#L64), something like:
> 
> ```
> {% if page.storybook_links and page.storybook_links.size > 0 %}
>     <ul class="m-list m-list--links">
>         {% for storybook_link in page.storybook_links %}
>             <li class="m-list__item">
>                 <a class="a-link a-link--jump" href="{{ storybook_link.href }}">
>                     [storybook icon goes here]
>                     <span class="a-link__text">{{ storybook_link.text }}</span>
>                 </a>
>             </li>
>         {% endfor %}
>     </ul>
> {% endif %}
> ```
> 
> Then we'd have a `storybook_links` field in the markdown files that can contain any number of links in it (the content of which would be the link text and the link URL). That would probably have to be hooked up to appear in Decap, too, right?
> 
> So it would be more upfront setup this way, but it would probably make it easier to add more links in the future (and would keep the markup in one spot if we ever need to change it).
> 
> What do you think?

Yeah, that is an interesting question. We probably could do that. I think if we did we would probably want to use the Web Component version (so `cfpb-link`). 

It would need to be hooked up in DeCap. 

One thing to point out is that I consciously was avoiding making changes to DeCap with this PR. That was/is because of our DeCap woes. I didn't want DeCap's issues to sidetrack this from getting done.  

Since right now DeCap's future is questionable, and there are bugs, I am not exactly sure what the right move would be for this. 

Thoughts?

niqjohnson:
@itsmedavep, I just pushed a couple of commits that make the Storybook links an included template. There are still some shortcomings with this:

1. As you discovered, if we want (a) a list (b) of links (c) that use an arbitrary icon on the left, our current web components don't seem to support that. The markup in the Storybook links include feels like the semantically correct structure, so I tried to use as many web components in it as worked (turns out it was just cfbp-icon).
2. I did make the Storybook links editable in Decap, but because we're now using an included template, the Decap preview doesn't render for those pages. I think I need to make a React component for the new include, just need to get that working.
3. Also apparently the Playwright tests failed? Not sure if that's a glitch or related to a change here, but I can investigate.

Anyway, given points 2 and 3, this probably isn't ready to merge yet, just wanted to push this to show what was in my head. Happy to revert back if you think this isn't worth the effort.

docs/_includes/storybook-links.html
```diff
@@ -0,0 +1,17 @@
+<ul class="m-list m-list--links u-mt20">
+    {% for storybook in include.storybooks %}
+    {% if include.scope == "variation" %}
+        {% assign storybook_type = storybook.variation_storybook_type %}
+        {% assign storybook_path = storybook.variation_storybook_path %}   
+    {% else %}
+        {% assign storybook_type = storybook.storybook_type %}
+        {% assign storybook_path = storybook.storybook_path %}
+    {% endif %}
+    <li class="m-list__item">
```
itsmedavep: If we want the Storybook icon and the link I think we can do it by having something like:

```html
<cfpb-icon-text icon-left="storybook" class="storybook-brand-color">
     <cfpb-link><a href="https://cfpb.github.io/{{ storybook_path }}">{{ storybook_type }}</a></cfpb-link>
</cfpb-icon-text>
```

itsmedavep:
@niqjohnson I added a comment directly to storybook-links.html that shows a way of using Web Components. 

I think if we want to get this stuff working using the Web Components in the preview we probably have do an import in `main-preview.js` for the `Cfpblink` and then initialize it. 

Then in `ReactLiquid.jsx` something like bundle every file under _includes so that Liquid `include` tags can be resolved by the browser. Keys would be paths relative to `_include` e.g. `'storybook-links.html'`

So some such like:

```jsx
const includeFiles = import.meta.glob(....), {
  query: '?raw',
  import: 'default',
  eager: true,
}

const templates = Object.fromEntries(
    Object.enteries(includeFiles).map(([path, source]) => [
    path.replace(...),
    source,
    ]);
    
const engine = new Liquid({
   dynamicPartials: false,
   jekyllInclude: true,
   templates,
});
```

niqjohnson:
@itsmedavep, I backtracked on having the Storybook links template be an include. It was adding way more Decap/LiquidJS overhead than it was worth, especially given that the markup for Storybook links needed to be slightly different depending on if it was page-level or variation-level, so it wasn't even a simple include anyway. I added the markup directly into `variation-content.html` along with a TODO that it could be pulled into an include someday. Now those links render great both on the page and in the preview. I also adopted your suggestion of using `cfpb-link` with `cfpb-icon-text`, and that's working nicely. I'm happy with this approach if you are.

niqjohnson (approved):

